Issue #355, opened 2025-11-01
benchmark CI workflow is broken


yegor256:
See: https://github.com/yegor256/factbase/actions/runs/18985728172/job/54228803667?pr=354

```
ArgumentError: comparison of Time with String failed (ArgumentError)
/home/runner/work/factbase/factbase/lib/factbase/indexed/indexed_term.rb:96:in 'Enumerable#sort_by'
/home/runner/work/factbase/factbase/lib/factbase/indexed/indexed_term.rb:96:in 'Array#sort_by!'
/home/runner/work/factbase/factbase/lib/factbase/indexed/indexed_term.rb:96:in 'Factbase::IndexedTerm#predict'
/home/runner/work/factbase/factbase/lib/factbase/query.rb:47:in 'Factbase::Query#each'
/home/runner/work/factbase/factbase/lib/factbase/cached/cached_query.rb:38:in 'Enumerator#each'
/home/runner/work/factbase/factbase/lib/factbase/cached/cached_query.rb:38:in 'Enumerable#to_a'
/home/runner/work/factbase/factbase/lib/factbase/cached/cached_query.rb:38:in 'Factbase::CachedQuery#each'
/home/runner/work/factbase/factbase/lib/factbase/sync/sync_query.rb:35:in 'block in Factbase::SyncQuery#each'
/home/runner/work/factbase/factbase/lib/factbase/sync/sync_query.rb:61:in 'Monitor#synchronize'
/home/runner/work/factbase/factbase/lib/factbase/sync/sync_query.rb:61:in 'Factbase::SyncQuery#try_lock'
/home/runner/work/factbase/factbase/lib/factbase/sync/sync_query.rb:34:in 'Factbase::SyncQuery#each'
/home/runner/work/factbase/factbase/benchmark/bench_query.rb:48:in 'Enumerator#each'
/home/runner/work/factbase/factbase/benchmark/bench_query.rb:48:in 'Enumerable#to_a'
/home/runner/work/factbase/factbase/benchmark/bench_query.rb:48:in 'block (4 levels) in Object#bench_query'
/home/runner/work/factbase/factbase/benchmark/bench_query.rb:47:in 'block (3 levels) in Object#bench_query'
/home/runner/work/factbase/factbase/benchmark/bench_query.rb:46:in 'block (2 levels) in Object#bench_query'
/home/runner/work/factbase/factbase/benchmark/bench_query.rb:45:in 'block in Object#bench_query'
[...]

PR #359, opened 2025-11-02
fix(#355): `bench_query` and `bench_lots_facts` conflict

volodya-lombrozo:
This PR fixes the problem with these benchmarks.

The Issue:

`bench_query` creates a few facts with `Time.now.iso8601` attributes which are strings.
`bench_lots_facts` creates a few facts with `Time.now` attributes which are time values.
All the facts are saved in the same shared factbase.
When `bench_query` acquired facts by using `(gt time '2024-...)`, it let to the type conflict.

This PR fixes this conflict

Related to #355


volodya-lombrozo:
@yegor256 could you have a look, please?

Issue #302, opened 2025-09-05
each term has to be in its own class


yegor256:
Current design, where we `include` many methods into the same `Term` class is ugly. Instead, we should create one class per term: `terms/or.rb`, `terms/unique.rb`, etc.

volodya-lombrozo:
@yegor256 could you assign it to me, please?

0pdd:
@yegor256 2 puzzles [#330](https://github.com/yegor256/factbase/issues/330), [#357](https://github.com/yegor256/factbase/issues/357) are still not solved.

PR #401, opened 2025-11-17
feat(#302): create separate classes for 'always' and 'never' terms

volodya-lombrozo:
This PR refactors the `Factbase::Term` class by introducing separate classes for the `always` and `never` terms.

Related to #302


volodya-lombrozo:
@yegor256 could you have a look, please?

PR #430, opened 2025-11-29
enhancement(#424): make txn lazy

### LINKED ISSUE #424: Transactions could be lazy
may help solve #245

Currently, in the `Factbase.txn()` we make a copy of the entire factbase and then let the block work with it. This may be expensive if the factbase is big. Would be more effective to make the `.txn()` method lazy. It should not make a copy but simply `yield` a decorated factbase that will do the copy only when/if an `insert()` or any other modifications happen to it. If no modifications, the copy won't be created.

HaidarJbeily7:
Closes #424 

### Solution

Implemented lazy copy-on-write for transactions. The copy is deferred until the first modification (insert, delete, or property change) actually occurs. If no modification happens, no copy is made.

### Changes

- Added `Factbase::LazyTaped` class that extends `Factbase::Taped`
- Modified `Factbase#txn` to use `LazyTaped` instead of copying upfront
- Added `bench_lazy_txn.rb` benchmark to measure transaction performance

### Benchmark Results

| Scenario | 50K Facts | 100K Facts |
|----------|-----------|------------|
| **Read-only txn** | 8.13s → 5.36s (**34% faster**) | 17.24s → 11.16s (**35% faster**) |
| **Rollback txn (no mods)** | 8.25s → 5.84s (**29% faster**) | 17.21s → 10.93s (**36% faster**) |
| Insert in txn | 3.55s → 3.79s (~same) | 10.22s → 10.36s (~same) |
| Modify in txn | 29.84s → 30.13s (~same) | 62.85s → 63.12s (~same) |

### DIFF
--- a/lib/factbase/lazy_taped.rb
+++ b/lib/factbase/lazy_taped.rb
@@ -1,0 +1,294 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'decoor'
+require_relative '../factbase'
+require_relative 'taped'
+
+# A lazy extension of Taped that defers copying of the original maps until the first
+# modification occurs. This optimizes read-only transactions and rollbacks
+# on large factbases by avoiding unnecessary deep copies.
+#
+# Author:: Yegor Bugayenko (yegor256@gmail.com)
+# Copyright:: Copyright (c) 2024-2025 Yegor Bugayenko
+# License:: MIT
+class Factbase::LazyTaped < Factbase::Taped
+  # Constructor.
+  # @param [Array] original The original maps to wrap
+  # @param [LazyTaped] root The root LazyTaped instance (nil if this is the root)
+  def initialize(original, root: nil)
+    @original = original
+    @root = root
+    if root.nil?
+      # This is the root instance - owns the lazy copy state
+      @copied = false
+      @pairs = {}
+      @original_to_copy = {}
+    end
+    super([], inserted: root_inserted, deleted: root_deleted, added: root_added)
+  end
+
+  # Returns true if the copy has been made (i.e., a modification occurred).
+  def copied?
+    root_instance.instance_variable_get(:@copied)
+  end
+
+  # Returns the pairs mapping (copy_oid -> original_oid) for reconciliation.
+  def pairs
+    root_instance.instance_variable_get(:@pairs)
+  end
+
+  def find_by_object_id(oid)
+    return nil unless copied?
+    @origin.find { |m| m.object_id == oid } ||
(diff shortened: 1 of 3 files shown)


yegor256:
@volodya-lombrozo please, review this one

lib/factbase/lazy_taped.rb
```diff
@@ -0,0 +1,294 @@
...
+
+require 'decoor'
+require_relative '../factbase'
+require_relative 'taped'
+
+# A lazy extension of Taped that defers copying of the original maps until the first
+# modification occurs. This optimizes read-only transactions and rollbacks
+# on large factbases by avoiding unnecessary deep copies.
+#
+# Author:: Yegor Bugayenko (yegor256@gmail.com)
```
volodya-lombrozo: @HaidarJbeily7 Are you the author?

lib/factbase/lazy_taped.rb
```diff
@@ -0,0 +1,294 @@
...
+  def root_deleted
+    root_instance.instance_variable_get(:@deleted) || @deleted || []
+  end
+
+  def root_added
+    root_instance.instance_variable_get(:@added) || @added || []
+  end
+
+  def sync_origin_from_root
+    # Find maps in root's @origin that correspond to our @original
```
volodya-lombrozo: @HaidarJbeily7 do we need this comment? If it is so, it's better to rewrite it to make it answer the question "why?"

https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/

volodya-lombrozo (changes requested):
@HaidarJbeily7 There is plenty of valuable code. Thank you! That said, I would highly recommend splitting this PR into several smaller ones. Doing so will benefit you in multiple ways:

1. You will receive more bonuses (currently, you might even face a penalty for such a large PR, despite its significant value). As I recall, we have a limit of approximately 130 lines of code.
2. The review process will be significantly faster.

[Helping others review your changes](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/helping-others-review-your-changes#write-small-pull-requests):

> Aim to create small, focused pull requests that fulfill a single purpose. Smaller pull requests are easier and faster to review and merge, leave less room to introduce bugs, and provide a clearer history of changes.

0crat:
@volodya-lombrozo Hey! Nice work on that review! 🎉 You just scored **+18 points** - here's the breakdown: **+18** base points for being a reviewer, **+5** for tackling those 363 hits-of-code (maxed out at 16 but still solid!), and **-5** since we only got 6 comments when 9+ would've been ideal. Your running total is now [+565](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - keep crushing it and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

PR #438, opened 2025-12-06
enhance(#424): make txns lazy

### LINKED ISSUE #424: Transactions could be lazy

HaidarJbeily7:
Closes #424 

## 50,000 Facts

| Operation | Before (real) | After (real) | Improvement |
|----------|---------------|--------------|-------------|
| read-only txn (no copy needed) | 7.6089s | 4.1905s | **45% faster** |
| rollback txn (no copy needed) | 7.4353s | 4.1346s | **44% faster** |
| insert in txn (copy triggered) | 3.0071s | 1.6199s | **46% faster (~1.85×)** |
| modify in txn (copy triggered) | 22.5860s | 21.2948s | **~6% faster** |

---

## 100,000 Facts

| Operation | Before (real) | After (real) | Improvement |
|----------|---------------|--------------|-------------|
| read-only txn (no copy needed) | 15.0183s | 9.2698s | **38% faster** |
| rollback txn (no copy needed) | 14.9860s | 9.1266s | **39% faster** |
| insert in txn (copy triggered) | 7.0102s | 4.1824s | **40% faster (~1.7×)** |
| modify in txn (copy triggered) | 59.6561s | 54.6600s | **~8% faster** |

---

# Raw Benchmark Output

### before
```
50000 facts: read-only txn (no copy needed) 7.608865
50000 facts: rollback txn (no copy needed) 7.435322
50000 facts: insert in txn (copy triggered) 3.007110
50000 facts: modify in txn (copy triggered) 22.586001
100000 facts: read-only txn (no copy needed) 15.018310
100000 facts: rollback txn (no copy needed) 14.986018
100000 facts: insert in txn (copy triggered) 7.010163
100000 facts: modify in txn (copy triggered) 59.656088
```

### after
```
50000 facts: read-only txn (no copy needed) 4.190509
50000 facts: rollback txn (no copy needed) 4.134652
50000 facts: insert in txn (copy triggered) 1.619944
50000 facts: modify in txn (copy triggered) 21.294831
100000 facts: read-only txn (no copy needed) 9.269808
100000 facts: rollback txn (no copy needed) 9.126606
100000 facts: insert in txn (copy triggered) 4.182432
100000 facts: modify in txn (copy triggered) 54.660025
```

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -25,7 +25,7 @@ jobs:
       - run: bundle install --no-color
       - run: bundle exec rake benchmark > stdout.txt
       - run: .github/publish-benchmark.sh
-      - uses: peter-evans/create-pull-request@v7
+      - uses: peter-evans/create-pull-request@v8
         with:
           sign-commits: true
           branch: benchmark-results
--- a/.github/workflows/markdown-lint.yml
+++ b/.github/workflows/markdown-lint.yml
@@ -20,4 +20,4 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v6
-      - uses: DavidAnson/markdownlint-cli2-action@v21.0.0
+      - uses: DavidAnson/markdownlint-cli2-action@v22.0.0


lib/factbase/lazy_taped.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  def pairs
+    return {} unless @pairs
+    result = {}.compare_by_identity
+    @pairs.each { |copied, original| result[copied] = original }
+    result
+  end
+
+  def inserted
```
volodya-lombrozo: @HaidarJbeily7 Could you add a comment which would answer the question "why do we need this method?"

lib/factbase/lazy_taped.rb
```diff
@@ -0,0 +1,131 @@
...
+    @inserted.uniq
+  end
+
+  def deleted
```
volodya-lombrozo: @HaidarJbeily7 Could you add a comment which would answer the question "why do we need this method?"

test/factbase/test_lazy_taped.rb
```diff
@@ -0,0 +1,252 @@
...
+    assert_equal(2, t.size)
+  end
+
+  def test_copies_on_property_set
+    original = [{ 'foo' => [1] }]
+    t = Factbase::LazyTaped.new(original)
+    t.each { |m| m['bar'] = 42 }
+    refute_empty(t.pairs)
+    assert_equal(1, t.added.size)
+    # Original should not be modified
```
volodya-lombrozo: @HaidarJbeily7 could you pass this text as an assertion message instead of using comments?

```ruby
assert_nil(original[0]['bar'], "Original should not be modified")
```

lib/factbase/lazy_taped.rb
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  # Returns the unique object IDs of maps that were modified (properties added).
+  # This is used during transaction commit to track the churn (number of changes).
+  def added
+    @added.uniq
+  end
+
+  def find_by_object_id(oid)
+    ensure_copied if @copied
```
volodya-lombrozo: @HaidarJbeily7 Sorry, I can't get why do we need to copy the entire factbase on a simple read operation? Maybe I'm missing something?

HaidarJbeily7: you're right, unneeded check, I will run the bench again as I removed the check

lib/factbase/lazy_taped.rb
```diff
@@ -0,0 +1,142 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative '../factbase'
+require_relative 'taped'
+require_relative 'lazy_taped_hash'
+
+# A lazy decorator of an Array with HashMaps that defers copying until modification.
+class Factbase::LazyTaped
+  def initialize(origin)
```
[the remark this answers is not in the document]
HaidarJbeily7: added

PR #450, opened 2025-12-12
bug: fix incremental indexing

HaidarJbeily7:
# Summary

This PR fixes several bugs in the incremental indexing implementation that caused incorrect query results when facts were inserted, deleted, or modified after initial indexing.

### indexed_and.rb / indexed_eq.rb

-   Fixed result collection logic
-   Replaced `reduce(&:|)` with `flat_map(...).uniq(&:object_id)` to  properly deduplicate facts
-   Properly wrap results in `Factbase::Taped` when the source tracks mutations (inserted/deleted/added)

### indexed_factbase.rb

-   Always wrap queries in `IndexedQuery`
-   Removed conditional that only wrapped "abstract" terms, ensuring
    consistent behavior for all queries

### indexed_query.rb

-   `one()`: Changed default params from `nil` to `{}` to avoid nil errors
-   `delete!`: Clear the index before returning (was clearing after return, causing stale index)

# Test Suite

Added `test_incremental_indexing.rb` with 9 test cases covering: 
- Query correctness after adding properties to existing facts
- Index clearing on delete and transaction rollback 
- Incremental indexing on sequential inserts 
- Delete-then-insert scenarios 
- NOT, GT, and AND terms after modifications

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -36,7 +36,7 @@ GEM
     logger (1.7.0)
     loog (0.6.1)
       logger (~> 1.0)
-    minitest (5.26.2)
+    minitest (5.27.0)
     minitest-reporters (1.7.1)
       ansi
       builder


yegor256:
@rultor release, tag is `0.19.1`

Issue #474, opened 2026-02-03
IndexedFactbase changes the core behavior of Factbase by returning inconsistent results


belfil:
A decorator should only affect performance, not the outcome. 
However, IndexedFactbase changes the core behavior of Factbase by returning inconsistent results.

Plain Factbase (Expected behavior):
```
fb = Factbase.new
f = fb.insert

f.status = 'new'
puts fb.query("(eq status 'open')").any? # => false

f.status = 'open'
puts fb.query("(eq status 'open')").any? # => true
```
IndexedFactbase (Buggy behavior):
```
fb = Factbase::IndexedFactbase.new(Factbase.new)
f = fb.insert

f.status = 'new'
puts fb.query("(eq status 'open')").any? # => false

f.status = 'open'
puts fb.query("(eq status 'open')").any? # => false
```

yegor256:
@belfil definitely a bug, let's fix it!

yegor256:
@rultor release, tag is `0.19.4`

PR #476, opened 2026-02-03
Fix stale index consistency via shared fresh Set

belfil:
Closes: https://github.com/yegor256/factbase/issues/474


belfil:
@yegor256 please check

PR #484, opened 2026-02-09
Remove unused variable and empty line

belfil:
Let's clean up some code smells
- An empty line.
- Unused local variable.

### DIFF
--- a/lib/factbase/indexed/indexed_and.rb
+++ b/lib/factbase/indexed/indexed_and.rb
@@ -12,7 +12,6 @@ class Factbase::IndexedAnd
 
   def predict(maps, fb, params)
     return nil if @idx.nil?
-    key = [maps.object_id, @term.operands.first, @term.op]
     r = nil
     if @term.operands.all? { |o| o.op == :eq } && @term.operands.size > 1 \
       && @term.operands.all? { |o| o.operands.first.is_a?(Symbol) && _scalar?(o.operands[1]) }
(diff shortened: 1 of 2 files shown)


belfil:
> @belfil how come this unused variable is not seen by Rubocop? This is the true problem which we should fix.

@yegor256  Even though
_This cop detects unreferenced reassignments in branches, loops, rescue/ensure, etc._
https://docs.rubocop.org/rubocop/cops_lint.html#lintuselessassignment
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/UselessAssignment

RuboCop catches **UselessAssignment** in the same scope, but ignores assignments overwritten inside conditionals:

```ruby
# frozen_string_literal: true

# SPDX-FileCopyrightText: Copyright (c) 2024-2026
# SPDX-License-Identifier: MIT

def skip(ok)
  val = 0 # RuboCop is silent here
  if ok
    val = 5
    val += 5 if ok
    p val
  end
  0
end

def catch(ok)
  num = 0 # RuboCop sees this
  if ok
    num = 5
    num += 5
    p num
  end
  0
end
```

```bash
$ bundle exec rubocop useless.rb

Inspecting 1 file
W

Offenses:

useless.rb:17:3: W: [Correctable] Lint/UselessAssignment: Useless assignment to variable - num.
  num = 0 # RuboCop sees this
  ^^^

1 file inspected, 1 offense detected, 1 offense autocorrectable
```

Issue #498, opened 2026-02-15
IndexedGt fails to find duplicates inside a transaction


belfil:
The IndexedGt term fails to correctly find facts with duplicate property values when they are inserted inside a transaction.

```ruby
def test_indexed_term_gt_keeps_duplicates
    fb = Factbase::IndexedFactbase.new(Factbase.new)
    fb.insert.v = 20
    fb.insert.v = 20
    fb.insert.v = 20
    assert_equal(3, fb.query('(gt v 10)').each.to_a.size
  end
```
```
 Expected: 3
 Actual: 1
```
```ruby
  def test_indexed_term_gt_keeps_duplicates_in_txn
    fb = Factbase::IndexedFactbase.new(Factbase.new)
    fb.txn do |fbt|
      fbt.insert.v = 20
      fbt.insert.v = 20
      fbt.insert.v = 20
      assert_equal(3, fbt.query('(gt v 10)').each.to_a.size)
    end
    assert_equal(3, fb.query('(gt v 10)').each.to_a.size)
  end
```
```
 Expected: 3
 Actual: 1
```

Issue #491, opened 2026-02-13
`Factbase::Fuzz`


yegor256:
Let's create a new class that can generate factbases with big amount of random data, for testing:

```
require 'factbase/fuzz'
fb = Factbase::Fuzz.make
ff = f.query('(always)').to_a
assert(ff.size > 100 && ff.size < 1000)
```

Such a class would help in tests, where we don't care about content, but just need a large factbase.

PR #499, opened 2026-02-15
#491 Fix loss of duplicates in IndexedGt and optimize search performance

belfil:
Closes: https://github.com/yegor256/factbase/issues/498

## Summary of Changes

This PR fixes a bug where `IndexedGt` failed to correctly find all facts with duplicate property values, especially when inserted inside a transaction.

## Changes
* **Fix**: Updated `IndexedGt` to correctly store and retrieve all facts, even if they share identical property values.
* **Integrity**: Added `uniq(&:object_id)` to the result mapping to ensure distinct facts are preserved while avoiding duplicate entries for a single fact with multiple matching values.
* **Lazy Indexing**: Refactored `_feed` to process only new facts since the last indexing cycle.
* **Decorator Support**: Ensured consistency for `Taped` and `LazyTaped` via the `repack` method.

### DIFF
--- a/lib/factbase/indexed/indexed_gt.rb
+++ b/lib/factbase/indexed/indexed_gt.rb
@@ -11,41 +11,39 @@ class Factbase::IndexedGt
   end
 
   def predict(maps, _fb, params)
-    return nil if @idx.nil?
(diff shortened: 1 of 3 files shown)


belfil:
@yegor256 I resolved the conflict, please check.

Issue #506, opened 2026-02-15
IndexedOne fails to find duplicates inside a transaction


belfil:
The IndexedOne term fails to correctly find facts with duplicate property values when they are inserted inside a transaction.

```ruby
def test_indexed_term_one_keeps_duplicates
    fb = Factbase::IndexedFactbase.new(Factbase.new)
    fb.insert.scope = 10
    fb.insert.scope = 10
    assert_equal(2, fb.query('(one scope)').each.to_a.size)
  end
```
```
 Expected: 2
 Actual: 1
```
```ruby
  def test_indexed_term_one_keeps_duplicates_in_txn
    fb = Factbase::IndexedFactbase.new(Factbase.new)
    fb.txn do |fbt|
      fbt.insert.scope = 10
      fbt.insert.scope = 10
      assert_equal(2, fbt.query('(one scope)').each.to_a.size)
    end
    assert_equal(2, fb.query('(one scope)').each.to_a.size)
  end
```
```
 Expected: 2
 Actual: 1
```

PR #507, opened 2026-02-15
#506: Fixed IndexedOne to preserve identity and transaction state

belfil:
Closes: https://github.com/yegor256/factbase/issues/506

## Summary of Changes

This PR fixes a bug where `IndexedOne` failed to correctly find all facts with duplicate property values, especially when inserted inside a transaction.

## Changes
* **Fix**: Updated `IndexedOne` to correctly store and retrieve all facts, even if they share identical property values.
* **Integrity**: Added `uniq(&:object_id)` to the result mapping to ensure distinct facts are preserved while avoiding duplicate entries for a single fact with multiple matching values.
* **Lazy Indexing**: Refactored `_feed` to process only new facts since the last indexing cycle.
* **Decorator Support**: Ensured consistency for `Taped` and `LazyTaped` via the `repack` method.

### DIFF
--- a/lib/factbase/indexed/indexed_one.rb
+++ b/lib/factbase/indexed/indexed_one.rb
@@ -11,25 +11,21 @@ class Factbase::IndexedOne
   end
 
   def predict(maps, _fb, _params)
-    return nil if @idx.nil?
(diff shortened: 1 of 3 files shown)


yegor256:
@rultor merge

Issue #488, opened 2026-02-12
Add benchmarks for indexed terms performance


belfil:
I propose adding a new set of benchmarks to track the performance of indexed queries compared to plain sequential scans. 
This will help us monitor indexing efficiency and catch regressions in query execution.

Example benchmark
```ruby
def bench_indexes_terms(bmk, _fb)
  # Total defines the scale of the dataset (number of facts).
  total = 15_000
  # Iterations (repeats) for each benchmark run.
  iterations = 5
  # Selectivity controls how many facts match the predicate.
  # For example, 10 means 10% of facts will satisfy the condition.
  selectivity = 10
  # Cardinality defines the diversity of "noise" data (non-matching facts).
  # High cardinality (e.g., 1000) makes matching facts stand out more,
  # while low cardinality (e.g., 2) creates many duplicates, forcing
  # the index to work harder on filtering.
  cardinality = 5
  cases = [
    { term: 'absent', query: '(absent status)',...},
    { term: 'exists', query: '(exists status)', },
    { term: 'eq',     query: '(eq status 1',...},
    { term: 'not',    query: '(not (eq status 1)',...},
    { term: 'gt',     query: '(gt score 500)',...},
    { term: 'lt',     query: '(lt score 500',...},
    { term: 'and',    query: '(and (eq status 1) (gt tag 1)',...},
    { term: 'one',    query: '(one tag)',...},
    { term: 'or',     query: '(or (eq status 1) (eq tag 1))',...},
    { term: 'unique', query: '(unique status tag)',...},
  ]
  ...
```

Example output
```bash
query 15k facts  sel: 10%  card: 5 plain absent                    0.339006   0.000000   0.339006 (  0.338999)
query 15k facts  sel: 10%  card: 5 absent indexed(cold)            0.084100   0.000001   0.084101 (  0.084099)
query 15k facts  sel: 10%  card: 5 absent indexed(warm)            0.361519   0.001995   0.363514 (  0.363513)
...
query 15k facts  sel: 10%  card: 5 plain unique                    1.084107   0.000000   1.084107 (  1.084100)
query 15k facts  sel: 10%  card: 5 unique indexed(cold)            0.056572   0.000000   0.056572 (  0.056582)
query 15k facts  sel: 10%  card: 5 unique indexed(warm)            0.152109   0.000000   0.152109 (  0.152107)
```

PR #516, opened 2026-02-17
#488: optimized indexing benchmark with strict cold/warm isolation

belfil:
https://github.com/yegor256/factbase/issues/488 The indexing benchmark was misleading. 
The "cold" test was only cold for the first iteration, while the other 5 repeats were using the cached index.

**Changes:**
- Fixed `indexed(cold)` to reset the index inside the loop (true cold start).
- Unified the number of repeats for `plain`, `cold`, and `warm` tests.
- Added explicit pre-warming for the `warm` test to isolate raw search speed.

Now the benchmark table shows the real "tax" of index construction.


belfil:
@yegor256 Please check

Issue #442, opened 2025-12-08
test_query.rb:349-362: Fix the flaky test...


0pdd:
The puzzle `330-d3182961` from #330 has to be resolved: 

https://github.com/yegor256/factbase/blob/f97a0dc06f268628e52bf120ccada737ef827f97/test/factbase/test_query.rb#L349-L362

The puzzle was created by @yegor256 on 08-Dec-25. 

Estimate: 90 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #529, opened 2026-02-25
#442: Fix flaky test

belfil:
Closes: https://github.com/yegor256/factbase/issues/442

The `test_txn_performance_degradation` test was flaky because it relied on execution time (`Time.now`), which is highly sensitive to CPU load, JIT "warmup", and system background processes.

### Changes:
- Switched from **time-based** measurements to **object allocation** counts using `GC.stat(:total_allocated_objects)`.
- Added a **warmup phase** to ensure internal state and string caches are initialized before measurements.
- Implemented a **two-window comparison** (left half vs. right half) to detect systematic growth.
- Added a `max_growth` threshold (5%) to allow minor acceptable variance while catching real regressions.
- Improved error messages to show the actual growth percentage and raw stats.

This approach is deterministic and provides a stable way to detect performance degradation across different environments.

### Validation:
The test successfully detects artificial degradation. 
For example, injecting redundant operations into `txn`:

```ruby
def txn
  15.times { insert.id = 10 } # Artificial degradation
  result = @origin.txn do |fbt|
    yield Factbase::IndexedFactbase.new(fbt, @idx, @fresh)
  end
  # ...
end
```
```bash
TestQuery
test_txn_performance_degradation                                FAIL (2.36s)
indexed+plain. Growth 5.64% Max 5%. Total objects: [27200, 27472, 27789, 28103, 28418, 28732, 29049, 29363, 29678, 30002].
Expected 5 to be > 5.64.
```

### DIFF
--- a/test/factbase/test_query.rb
+++ b/test/factbase/test_query.rb
@@ -346,41 +346,31 @@ class TestQuery < Factbase::Test
     assert_includes(f.all_properties, 'foo')
   end
 
-  # @todo #330:90min Fix the flaky test 'test_txn_performance_degradation'.
-  # The 'test_txn_performance_degradation' sometimes fails, sometimes passes.
-  # The reason is not clear yet.
-  # I got the following error message:
-  # ```
-  #  In indexed+cached+rules+plain, the trend is up: 2.483, 2.567, 2.872,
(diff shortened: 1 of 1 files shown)


belfil:
@yegor256 Please review.

yegor256:
@belfil thanks!

Issue #537, opened 2026-02-28
Transaction crashes when query result size exceeds 128


belfil:
The transaction fails with NoMethodError: undefined method 'to_set' when executing a query that returns more than 128 facts.

```ruby
def test_indexed_term_and_keeps_many_duplicates_facts
    fb = Factbase::IndexedFactbase.new(Factbase.new)
    total = 130
    total.times do
      f = fb.insert
      f.cost = 'cost'
      f.scope = 'scope'
    end
    q = '(and (exists cost) (exists scope))'
    assert_equal(total, fb.query(q).each.to_a.size)
    fb.txn { |fbt| assert_equal(total, fbt.query(q).each.to_a.size) }
  end
```

```
TestIndexedFactbase
  test_indexed_term_and_keeps_many_duplicates_facts              ERROR (0.01s)
  Minitest::UnexpectedError:         NoMethodError: undefined method 'to_set' for an instance of Factbase::Taped
  lib/factbase/indexed/indexed_and.rb:54:in 'block in Factbase::IndexedAnd#predict'
  lib/factbase/indexed/indexed_and.rb:48:in 'Array#each'
  lib/factbase/indexed/indexed_and.rb:48:in 'Factbase::IndexedAnd#predict'
```

belfil:
@yegor256 Merged but rejected? @0crat is a tough judge! 🤠

PR #538, opened 2026-02-28
537: Fix and optimize intersection in IndexedAnd

belfil:
Closes: https://github.com/yegor256/factbase/issues/537
### Description
The `IndexedAnd` optimization (triggered for 128+ facts) fails during transactions because it expects an `Array` but receives a `Factbase::LazyTaped` or `Taped` decorator, causing `NoMethodError: undefined method 'to_set'`.

### Changes
- **Fix**: Added `.to_a` before building `Set` in `IndexedAnd` to support decorators.
- **Performance**: Optimized intersection by building the `Set` index from the **smaller** collection.


belfil:
@yegor256 Please check.

Issue #439, opened 2025-12-07
Memory leak in IndexedFactbase: stale index entries accumulate during transactions


yegor256:
## Summary

The `@idx` hash in `IndexedFactbase` accumulates stale entries during transactions, causing unbounded memory growth and exponential performance degradation over time.

## Root Cause

The index key in `indexed_eq.rb` and `indexed_and.rb` uses `maps.object_id`:

```ruby
# indexed_and.rb:20
key = [maps.object_id, props, :multi_and_eq]

# indexed_eq.rb:15
key = [maps.object_id, @term.operands.first, @term.op]
```

However, during transactions, new map copies are created with **new object_ids**:

```ruby
# factbase.rb:168-175
before =
  @maps.map do |m|
    n = m.transform_values(&:dup)
    pairs[n.object_id] = m.object_id
    n
  end
```

And `IndexedFactbase.txn` shares the **same `@idx`** hash:

```ruby
# indexed_factbase.rb:59-63
def txn
  @origin.txn do |fbt|
    yield Factbase::IndexedFactbase.new(fbt, @idx)  # same @idx!
  end
end
```

## The Problem

1. Each transaction creates new map copies with new `object_id`s
2. Queries inside transactions build new index entries with these new `object_id`s
3. Old index entries (with old `object_id`s) are **never removed**
4. `@idx.clear` only happens on `insert` or property assignment, not after transactions complete
5. The `@idx` hash grows unboundedly, storing references to old map copies

## Reproduction Scenario

```ruby
fb = Factbase::IndexedFactbase.new(Factbase.new)

# Add some initial facts
100.times do |i|
  f = fb.insert
  f.issue = i
  f.repository = 1
  f.what = 'issue-was-opened'
  f.where = 'github'
end

# Each transaction creates stale index entries
1000.times do
  fb.txn do |fbt|
    # This query builds index entries with transactional maps.object_id
    fbt.query("(and (eq issue 50) (eq repository 1) (eq what 'issue-was-opened') (eq where 'github'))").each.first
  end
  # After transaction completes, index entries become stale but are never cleaned
end

# @idx now contains 1000+ stale entries, each holding references to old map copies
```

## Impact

- **Memory leak**: Old map copies cannot be garbage collected (referenced by stale index entries)
- **Performance degradation**: Hash operations slow down as `@idx` grows
- **GC pressure**: Large heap triggers expensive GC cycles

[...]

Issue #321, opened 2025-10-05
a simple query takes 50+ seconds


yegor256:
Found it here: https://github.com/zerocracy/judges-action/actions/runs/18255591064/job/51976393568

```
E: RuntimeError: each() timed out after 60s (execution expired), fb size is 11422: (and (eq issue 85) (eq repository 802720676) (eq what 'issue-was-opened') (eq where 'github'))
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/impatient.rb:89:in 'Factbase::Impatient::Query#impatient'
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/impatient.rb:60:in 'Factbase::Impatient::Query#each'
   	/usr/local/bundle/gems/fbe-0.39.0/lib/fbe/if_absent.rb:75:in 'Fbe.if_absent'
   	/action/judges/find-all-issues/find-all-issues.rb:58:in 'block (6 levels) in <top (required)>'
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/impatient.rb:39:in 'block in Factbase::Impatient#txn'
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/logged.rb:55:in 'block in Factbase::Logged#txn'
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/sync/sync_factbase.rb:54:in 'block in Factbase::SyncFactbase#txn'
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/indexed/indexed_factbase.rb:61:in 'block in Factbase::IndexedFactbase#txn'
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/cached/cached_factbase.rb:61:in 'block in Factbase::CachedFactbase#txn'
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/pre.rb:42:in 'block in Factbase::Pre#txn'
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/rules.rb:54:in 'block in Factbase::Rules#txn'
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/tallied.rb:43:in 'block (2 levels) in Factbase::Tallied#txn'
   	/usr/local/bundle/gems/factbase-0.16.7/lib/factbase/tallied.rb:42:in 'Kernel#catch'
[...]

PR #560, opened 2026-04-29
#321: keep index warm across insert-only transactions

morphqdd:
## Problem

`IndexedFactbase#txn` called `@idx.clear` unconditionally after every transaction (added in #439), discarding the warm index even when only inserts occurred. The next query rebuilt the index from scratch over all facts — causing the 50+ second queries reported in #321.

## Root Cause

The `@idx.clear` in #439 was added to prevent unbounded index growth: keys are built on `maps.object_id`, and the inner `LazyTaped` inside `txn` has a different `object_id` than the outer `@maps`, so stale entries accumulated. The fix was correct but too aggressive — it also wiped valid entries for the outer factbase.

## Fix

Two changes in `IndexedFactbase#txn`:

1. **Isolated `inner_idx`** — the transaction now uses its own throwaway index instead of `@idx`. After the transaction it is simply discarded, eliminating the stale-key problem from #439 without touching the outer index.

2. **Conditional clear** — `@idx` is cleared only when existing facts were deleted or modified (`result.deleted > 0 || result.added > 0`), because in those cases the index holds stale object references. Insert-only transactions leave all existing maps untouched, so the index remains valid and incremental indexing picks up new facts automatically.

## Benchmark

100k facts, 10 cycles (`bundle exec rake benchmark[bench_indexed_txn,10]`):

```
100k facts: query after insert txn    0.013s  (~1ms/cycle)
100k facts: query after delete txn    0.006s
```

Before the fix, each cycle required a full index rebuild: ~450ms/cycle at 50k facts.

## Test

`test_transaction_clears_index` was rewritten as `test_transaction_clears_index_on_delete` — the correct trigger for clearing is a delete inside a committed transaction, not a rolled-back insert. A rollback leaves the factbase unchanged, so the index remains valid.

### DIFF
--- a/benchmark/bench_indexed_txn.rb
+++ b/benchmark/bench_indexed_txn.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require_relative '../lib/factbase'
+require_relative '../lib/factbase/indexed/indexed_factbase'
+
+# Benchmark for issue #321: index must survive insert-only transactions.
+# With IndexedFactbase wrapping, repeated queries after insert-only txns
+# should reuse the warm index rather than rebuild it from scratch each time.
+#
+# To run: bundle exec rake benchmark[bench_indexed_txn]
+def bench_indexed_txn(bmk, _fb, cycles)
+  total = 100_000
+  [total].each do |n|
+    p_total = "#{n / 1_000}k"
+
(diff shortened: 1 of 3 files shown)


morphqdd:

0crat:
@morphqdd Great work on your contribution! 🎉 You've earned **+24 points** and your running score is now [+202](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the momentum going and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #590, opened 2026-05-26
#576: handle invalid matches regexps

### LINKED ISSUE #576: `Matches` term leaks raw `RegexpError` on invalid pattern and recompiles regex per fact
## What happens

In [`lib/factbase/terms/matches.rb`](https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/matches.rb#L32), the `(matches foo "<pattern>")` term passes the user-supplied pattern straight into `String#match?` for every fact:

```ruby
str[0].to_s.match?(re[0])
```

Two defects on this one line.

**1. Raw `RegexpError` leaks to query author.** If the pattern is invalid (`"[a-"`, `"(unbalanced"`, etc.), Ruby's `String#match?` coerces the second argument to a `Regexp` at call time and raises `RegexpError: premature end of char-class` (or similar). The exception propagates out of the term unwrapped, so a query author sees `RegexpError` instead of a typed factbase error identifying the offending term and value. The neighboring `Compare` and `ToTime` terms have the same flavour of issue, but `matches` is the easiest to trip because users frequently iterate on patterns.

**2. Regex is recompiled per fact.** Because `re[0]` is a plain `String`, every `match?` call re-parses it. On a query that touches N facts, the same pattern is compiled N times. For large factbases this is a measurable hot spot.

Existing tests in `test/factbase/terms/test_matches.rb` only exercise valid, simple patterns, so neither defect surfaces in CI.

## What should happen

Compile the regex once when the term is constructed (or memoize on first evaluation), and wrap compilation in a typed error:

```ruby
def initialize(fb, operands)
  super
  begin
    @re = Regexp.new(operands[1].to_s)
  rescue RegexpError => e
    raise "Invalid regex in (matches ...): #{e.message}"
  end
end

def _matches(_fact, str)
  str[0].to_s.match?(@re)
end
```

Result: a malformed pattern produces a single, contextual error pointing the query author at the bad regex, and well-formed patterns run with one compilation instead of N.

GHX5T-SOL:
Fixes #576.

Changes:
- compile `Matches` regexp operands through a cached helper keyed by pattern
- wrap invalid regexp patterns with a contextual `RuntimeError`
- cover literal, property-sourced, cached, and invalid regexp cases

Validation:
- `PICKS=yes mise x ruby@3.3.11 -- bundle _2.6.8_ exec ruby test/factbase/terms/test_matches.rb` -> 4 tests, 11 assertions, 0 failures, 0 errors, 0 skips
- `mise x ruby@3.3.11 -- bundle _2.6.8_ exec rubocop lib/factbase/terms/matches.rb test/factbase/terms/test_matches.rb` -> 2 files inspected, no offenses detected
- `mise x ruby@3.3.11 -- bundle _2.6.8_ exec rake test` -> 516 tests, 29426 assertions, 1 failure, 0 errors, 2 skips; the failure is `TestLogged#test_proper_logging` and was reproduced on clean `master` with `PICKS=yes mise x ruby@3.3.11 -- bundle _2.6.8_ exec ruby test/factbase/test_logged.rb -n test_proper_logging`

### DIFF
--- a/lib/factbase/terms/matches.rb
+++ b/lib/factbase/terms/matches.rb
@@ -14,6 +14,7 @@ class Factbase::Matches < Factbase::TermBase
   def initialize(operands)
     super()
     @operands = operands
+    @regexps = {}
   end
 
   # Evaluate term on a fact.
@@ -29,6 +30,15 @@ class Factbase::Matches < Factbase::TermBase
     re = _values(1, fact, maps, fb)
     raise 'Regexp is nil' if re.nil?
     raise 'Exactly one regexp is expected' unless re.size == 1
-    str[0].to_s.match?(re[0])
+    str[0].to_s.match?(regexp(re[0]))
+  end
+
+  private
+
+  def regexp(pattern)
+    key = pattern.to_s
+    @regexps[key] ||= Regexp.new(key)
+  rescue RegexpError => e
+    raise "Invalid regexp '#{key}': #{e.message}"
   end
 end
--- a/test/factbase/terms/test_matches.rb
+++ b/test/factbase/terms/test_matches.rb
@@ -19,4 +19,26 @@ class TestMatches < Factbase::Test
     assert(t.evaluate(fact('foo' => 'hello 42'), [], Factbase.new))
     refute(t.evaluate(fact('foo' => 42), [], Factbase.new))
   end
+
+  def test_regexp_from_property
+    t = Factbase::Matches.new(%i[foo pattern])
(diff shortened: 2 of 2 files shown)


0crat:
@ghx5t-sol Great work on your contribution! 🎉 You've earned **+24 points** and your total score has reached **[+668](https://zerocracy.github.io/judges-action/zerocracy-vitals.html)** - excellent progress! [Keep delivering quality work](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) and don't forget to monitor your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #591, opened 2026-05-26
#579: wrap sprintf format errors

### LINKED ISSUE #579: `Sprintf` term leaks raw `ArgumentError` on format/value mismatch
## What happens

In [`lib/factbase/terms/sprintf.rb`](https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/sprintf.rb#L27), the `(sprintf "<fmt>" args...)` term invokes Ruby's `Kernel#format` directly, with no validation of operand types against the format string:

```ruby
format(*([fmt] + ops))
```

If `fmt` specifies a numeric placeholder but the operand is a non-numeric string — `(sprintf "%d" "hello")` — Ruby raises `ArgumentError: invalid value for Integer(): "hello"`. If `fmt` has more placeholders than the supplied arg count, Ruby raises `ArgumentError: too few arguments`. Both errors propagate out of the term unwrapped, so the query author sees a raw `ArgumentError` with no indication of which term or which fact caused it.

The existing test in `test/factbase/terms/test_sprintf.rb` only exercises matching format/value pairs, so the mismatch path is not covered.

## What should happen

Wrap the call in a typed error that names the term, the format string, and the operands:

```ruby
def _sprintf(_fact, ops)
  fmt = ops[0].to_s
  format(*([fmt] + ops[1..]))
rescue ArgumentError => e
  raise "Cannot format #{ops[1..].inspect} with '#{fmt}' in (sprintf ...): #{e.message}"
end
```

Result: a format/value mismatch surfaces as a single contextual error identifying the offending term, instead of a raw `ArgumentError` leaking from `Kernel#format`. This matches the wrapping pattern that should also be applied to the neighboring `to_time`, `matches`, and `compare` terms.

GHX5T-SOL:
Closes #579

Summary:
- wrap `Kernel#format` `ArgumentError` failures from `(sprintf ...)` with the term, format string, operands, and original message
- add regressions for an invalid numeric operand and too few format arguments

Validation:
- `PICKS=yes mise x ruby@3.3.11 -- bundle _2.6.8_ exec ruby test/factbase/terms/test_sprintf.rb` -> `3 tests, 11 assertions, 0 failures, 0 errors, 0 skips`
- `mise x ruby@3.3.11 -- bundle _2.6.8_ exec rubocop lib/factbase/terms/sprintf.rb test/factbase/terms/test_sprintf.rb` -> `2 files inspected, no offenses detected`

Full-suite limitation:
- `mise x ruby@3.3.11 -- bundle _2.6.8_ exec rake test` reached `515 tests, 29428 assertions, 1 failures, 0 errors, 2 skips`; the only failure was `TestLogged#test_proper_logging`, the same unrelated logging assertion failure previously reproduced on clean `master` with this Ruby/Bundler toolchain.

### DIFF
--- a/lib/factbase/terms/sprintf.rb
+++ b/lib/factbase/terms/sprintf.rb
@@ -24,6 +24,14 @@ class Factbase::Sprintf < Factbase::TermBase
   def evaluate(fact, maps, fb)
     fmt = _values(0, fact, maps, fb)[0]
     ops = (1..(@operands.length - 1)).map { |i| _values(i, fact, maps, fb)&.first }
+    formatted(fmt, ops)
+  end
+
+  private
+
+  def formatted(fmt, ops)
     format(*([fmt] + ops))
+  rescue ArgumentError => e
+    raise "Cannot format #{ops.inspect} with '#{fmt}' in (sprintf ...): #{e.message}"
   end
 end
--- a/test/factbase/terms/test_sprintf.rb
+++ b/test/factbase/terms/test_sprintf.rb
@@ -17,4 +17,24 @@ class TestSprintf < Factbase::Test
     t = Factbase::Sprintf.new(['hi, %s!', 'Jeff'])
     assert_equal('hi, Jeff!', t.evaluate(fact, [], Factbase.new))
   end
+
+  def test_rejects_invalid_format_operand
+    t = Factbase::Sprintf.new(['%d', 'hello'])
+    e =
+      assert_raises(RuntimeError) do
+        t.evaluate(fact, [], Factbase.new)
+      end
+    assert_includes(e.message, "Cannot format [\"hello\"] with '%d' in (sprintf ...):")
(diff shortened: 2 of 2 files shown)


0crat:
@ghx5t-sol Thanks for the contribution! You've earned **+20 points** for this: +24 as a basis; -4 for contributing less than 30 hits-of-code (you had 28). According to our [work policy](https://www.yegor256.com/2014/11/14/hits-of-code.html), contributions under 30 hits-of-code receive this deduction to encourage more substantial changes. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming with slightly larger contributions to maximize your bonus! Your running score is [+644](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #593, opened 2026-05-26
#578: wrap compare argument errors

### LINKED ISSUE #578: `Compare` term leaks raw `ArgumentError` on incompatible type comparisons
## What happens

In [`lib/factbase/terms/compare.rb`](https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/compare.rb#L34), the comparison terms `(gt a b)`, `(lt a b)`, `(gte a b)`, `(lte a b)` call `Comparable#<=>` family operators directly with no type guard:

```ruby
l.send(@op, r)
```

If the two operands are of incompatible types — `(gt 42 "hello")`, `(lt true 1.5)`, `(gt Time.now "yesterday")` — Ruby raises `ArgumentError: comparison of Integer with String failed`. The exception leaks unwrapped to the query layer.

Compare with the sibling `:==` path on lines 31-33 which explicitly floors `Time` values for graceful equality:

```ruby
return l.floor == r.floor if l.is_a?(Time) && r.is_a?(Time)
return l == r
```

So the design clearly cares about graceful semantics for equality but forgot to extend the care to ordering. The end result is that `(eq foo 'bar')` against an integer fact returns `false` gracefully, while `(gt foo 'bar')` against the same fact crashes the query.

## What should happen

Either return `false` for incompatible types (matching how `:==` already degrades gracefully), or wrap the `ArgumentError` in a typed factbase error that names the term and the operand classes:

```ruby
begin
  l.send(@op, r)
rescue ArgumentError => e
  raise "Cannot compare #{l.class} to #{r.class} in (#{@op} ...): #{e.message}"
end
```

Result: a mixed-type comparison either yields a deterministic `false` or surfaces a single contextual error, instead of a raw `ArgumentError` leaking out of the term layer.

GHX5T-SOL:
Closes #578

Summary:
- wrap `Compare` `ArgumentError` failures with the operator, operand values/classes, and original message
- add regressions for integer/string and time/string ordering comparisons

Validation:
- `PICKS=yes BUNDLE_PATH=vendor/bundle bundle _2.6.8_ exec ruby test/factbase/terms/test_compare.rb` -> `4 tests, 16 assertions, 0 failures, 0 errors, 0 skips`
- `BUNDLE_PATH=vendor/bundle bundle _2.6.8_ exec rubocop lib/factbase/terms/compare.rb test/factbase/terms/test_compare.rb` -> `2 files inspected, no offenses detected`
- `BUNDLE_PATH=vendor/bundle bundle _2.6.8_ exec rake test` -> `515 tests, 29436 assertions, 0 failures, 0 errors, 2 skips`

Environment note:
- Local validation used Ruby 3.3.8 with Bundler 2.6.8 on Kali; CI will cover the repository Ruby 3.4 matrix.

### DIFF
--- a/lib/factbase/terms/compare.rb
+++ b/lib/factbase/terms/compare.rb
@@ -31,8 +31,23 @@ class Factbase::Compare < Factbase::TermBase
       l = l.floor if l.is_a?(Time) && @op == :==
       rights.any? do |r|
         r = r.floor if r.is_a?(Time) && @op == :==
-        l.send(@op, r)
+        _compare(l, r)
       end
     end
+  end
+
+  private
+
+  # Compare values with a contextual error if Ruby rejects the operands.
+  # @param [Object] left Left value
+  # @param [Object] right Right value
+  # @return [Boolean] The result of the comparison
+  def _compare(left, right)
+    left.send(@op, right)
+  rescue ArgumentError => e
+    raise(
+      "Cannot compare #{left.inspect} (#{left.class}) " \
+      "with #{right.inspect} (#{right.class}) using (compare #{@op}): #{e.message}"
+    )
   end
 end
--- a/test/factbase/terms/test_compare.rb
+++ b/test/factbase/terms/test_compare.rb
@@ -18,4 +18,20 @@ class TestCompare < Factbase::Test
     t = Factbase::Compare.new(:<, [4, 2])
     refute(t.evaluate(fact, [], Factbase.new), 'Expected 2 < 4 to be true')
   end
+
+  def test_wraps_incompatible_type_comparison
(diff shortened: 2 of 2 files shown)


0crat:
@ghx5t-sol Hey! Nice work on your contribution – you just scored the full **24 points** 🎉 The code review process worked perfectly, and your contribution size hit that sweet spot in our [bonus calculation system](https://www.yegor256.com/2014/11/14/hits-of-code.html). Your running score is now [+600](https://zerocracy-vitals.html) – [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) these quality contributions coming and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

PR #602, opened 2026-06-05
docs: drop #599 incorrect "thread-safe" parenthetical in README

### LINKED ISSUE #599: README.md:22 claims `Factbase` is thread-safe, contradicting README.md:285 and lib/factbase.rb:78
The opening example in `README.md:22` opens with "Here is how you use it (it's thread-safe, by the way):", but the architecture section in `README.md:280-287` says the opposite: "The bare `Factbase` class is not thread-safe; new programmers must wrap it with `SyncFactbase` before sharing across threads."

The class header in `lib/factbase.rb:78` agrees with the second claim and states "It is NOT thread-safe!". A reader who copies the opening snippet will assume `Factbase.new` is safe to share across threads, miss the `SyncFactbase` decorator the architecture section points at, and hit data races the decorator exists to prevent.

The fix is to drop the parenthetical on line 22, and, if the thread-safety property is worth surfacing that early in the document, replace it with a one-line note pointing at `Factbase::SyncFactbase` for shared access.

kreinba:
The opening example on `README.md:22` told readers the gem is thread-safe in passing, but the architecture section on `README.md:285-287` and the class header on `lib/factbase.rb:78` both state that the bare `Factbase` is not thread-safe and must be wrapped with `Factbase::SyncFactbase` before being shared across threads. A reader following the opening snippet would assume `Factbase.new` is safe to share, miss the `SyncFactbase` decorator the architecture section points at, and hit data races the decorator exists to prevent.

The patch replaces the parenthetical so the opening example agrees with the rest of the document and surfaces the `SyncFactbase` decorator at the very first usage example. Behaviour, public API, and the persisted format are untouched.

Verified locally with `npx markdownlint-cli2 README.md` (0 errors). No code paths changed, so the rake test suite and rubocop runs are unaffected by the diff.

Closes #599

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,8 @@ The values are either atomic literals or non-empty sets of literals.
 It is possible to delete a fact, but impossible to delete a property
 from a fact.
 
-Here is how you use it (it's thread-safe, by the way):
+Here is how you use it (wrap it with `Factbase::SyncFactbase`
+for thread-safe access):
 
 ```ruby
 fb = Factbase.new


0crat:
@kreinba **Great work!** 🎉 You've earned **+20 points** (+24 base, -4 for 3 hits-of-code). [Keep the momentum going](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) - you're at [+1594 total points](https://zerocracy.github.io/judges-action/zerocracy-vitals.html)! Check your [Zerocracy account](https://www.zerocracy.com/) for updates.

Issue #632, opened 2026-06-10
Missing test coverage for defn/undef operators


VasilevNStas:
Add tests for edge cases in defn and undef: bad names (digits, caps), non-symbol arguments, eval execution, undef of nonexistent method.

VasilevNStas:
Fixed in PR #

VasilevNStas:
Fixed in #635

PR #635, opened 2026-06-10
#632: Add test coverage for defn/undef operators

VasilevNStas:
Add missing test coverage for defn and undef operators:

- defn with bad names (digits, caps)
- defn with non-symbol argument
- defn eval arbitrary Ruby code
- defn then undef then use raises RuntimeError
- undef nonexistent method
- undef with non-symbol argument

### DIFF
--- a/test/factbase/terms/test_defn.rb
+++ b/test/factbase/terms/test_defn.rb
@@ -2,6 +2,7 @@
 
 require_relative '../../../lib/factbase/term'
 require_relative '../../../lib/factbase/terms/defn'
+require_relative '../../../lib/factbase/terms/undef'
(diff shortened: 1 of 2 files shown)


0crat:
@vasilevnstas Hello! I noticed that your Git branch name "`defn-undef-tests`" doesn't follow our [naming convention](https://www.yegor256.com/2014/04/15/github-guidelines.html). According to our team policy, this results in a **-6 point** deduction. Please use the ticket number as your branch name next time - in this case, "`634`" would be the correct format. Your running score is [+1066](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); you can also check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #640, opened 2026-06-10
#640: Add IndexedGte, IndexedLte; fix IndexedAnd/Unique predict

VasilevNStas:
ARCH changes:

1. **IndexedGte/IndexedLte** — new indexed terms for >= and <=, using bsearch on sorted value arrays. Same pattern as IndexedGt/IndexedLt.
2. **IndexedAnd#predict** — return nil when any operand can't be predicted (instead of returning partial result from previous operands). Prevents incorrect early-exit optimization.
3. **IndexedUnique#predict** — return nil (unique is stateful and can't be index-predicted).

### DIFF
--- a/lib/factbase/indexed/indexed_and.rb
+++ b/lib/factbase/indexed/indexed_and.rb
@@ -45,9 +45,13 @@ class Factbase::IndexedAnd
       j = tuples.flat_map { |t| entry[:index][t] || [] }.uniq(&:object_id)
       r = maps.respond_to?(:repack) ? maps.repack(j) : j
     else
+      fail = false
       @term.operands.each do |o|
         n = o.predict(maps, fb, params)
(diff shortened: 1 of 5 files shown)


0crat:
@vasilevnstas Hey! Just a heads up - naming your Git branch `indexed-gte-lte` doesn't follow our [branch naming convention](https://www.yegor256.com/2014/04/15/github-guidelines.html), so you've got a **-6 point** penalty this time. The policy is pretty straightforward - just use the ticket number as your branch name, so `639` would've been perfect here! Your running score is [+1030](https://zerocracy.github.io/judges-action/zerocracy-vitals.html), so you're still doing great overall. Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) and keep up the good work! 🚀

PR #641, opened 2026-06-10
#601: Syntax#to_tokens fuses identifiers separated by newline or tab

### LINKED ISSUE #601: lib/factbase/syntax.rb:120: `to_tokens` fuses identifier tokens separated by `\t`, `\n`, or `\r` without a trailing space
`Factbase::Syntax#to_tokens` in `lib/factbase/syntax.rb` around line 120 keeps a `spaces = [' ', ')']` list that decides when to flush the running token accumulator `acc` into the token list. The `case c` block a few lines below also treats `"\n"`, `"\t"`, and `"\r"` as whitespace and skips them with `next`, but those three characters never trigger the flush at line 120 because they are not in `spaces`. The result is that any tab or bare newline (one not followed by a space) sitting directly between two identifier characters silently glues the two tokens together.

A query like `(eq foo\n1)` tokenises to `[:open, "eq", "foo1", :close]` instead of `[:open, "eq", "foo", 1, :close]`, and `(eq\tfoo 1)` tokenises to `[:open, "eqfoo", 1, :close]`. The existing tests happen to dodge the bug because every tab or newline in `test/factbase/test_syntax.rb` is followed by a space or by `)` before the next token, so the accumulator is already empty by the time the tab or newline arrives. Once a user formats a query across lines without indenting with spaces, or pastes a query that contains a tab, the parser silently builds the wrong AST and the downstream `Factbase::Term` either raises `ArgumentError: Wrong symbol format` for the fused identifier or, worse, matches the wrong property.

The smallest fix is to extend `spaces` to `[' ', ')', "\n", "\t", "\r"]` so the accumulator is flushed on every whitespace character the `case c` branch already swallows, and to add a regression test that asserts `(eq foo\n1)` and `(eq\tfoo 1)` produce the same term as `(eq foo 1)`.

VasilevNStas:
Bug fix for #601.

**Problem:** `Syntax#to_tokens` keeps `spaces = [' ', ')']` for flushing the token accumulator. Tabs and newlines (`\n`, `\t`, `\r`) are skipped by the `case c` branch but never trigger a flush, fusing adjacent identifiers.

For example:
- `(eq foo\n1)` tokenises to `[:open, "eq", "foo1", :close]` instead of `[:open, "eq", "foo", 1, :close]`
- `(eq\tfoo 1)` tokenises to `[:open, "eqfoo", 1, :close]`

**Fix:** Added `\n`, `\t`, `\r` to the `spaces` list.

**Tests:** Added `test_tokenizes_across_newline` and `test_tokenizes_across_tab` regression tests.

Fixes #601

### DIFF
--- a/lib/factbase/syntax.rb
+++ b/lib/factbase/syntax.rb
@@ -99,7 +99,7 @@ class Factbase::Syntax
     list = []
     acc = ''
     quotes = ['\'', '"']
-    spaces = [' ', ')']
+    spaces = [' ', ')', "\n", "\t", "\r"]
     string = false
     comment = false
     @query.to_s.chars.each do |c|
--- a/test/factbase/test_syntax.rb
+++ b/test/factbase/test_syntax.rb
@@ -57,6 +57,14 @@ class TestSyntax < Factbase::Test
     end
   end
 
+  def test_tokenizes_across_newline
+    assert_equal('(eq foo 1)', Factbase::Syntax.new("(eq foo\n1)").to_term.to_s)
+  end
+
+  def test_tokenizes_across_tab
+    assert_equal('(eq foo 1)', Factbase::Syntax.new("(eq\tfoo 1)").to_term.to_s)
+  end
+
   def test_exact_parsing
     [
       '(foo)',


0crat:
@vasilevnstas Great work on your contribution! 🎉 You've earned **+20 points** (+24 base, -4 for 10 hits-of-code). [Keep the momentum going](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) with quality contributions! Your total score is now [+1766](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #646, opened 2026-06-14
#485: detect useless pre-conditional assignments via test

### LINKED ISSUE #485: Detect variables overwritten inside conditionals (UselessAssignment)
Rubocop’s Lint/UselessAssignment doe's not detect variables that are overwritten inside conditional blocks, even though the initial assignment is never used. This makes it easy to miss truly useless assignments in our code.

**Example**
```ruby
# frozen_string_literal: true

# SPDX-FileCopyrightText: Copyright (c) 2024-2026
# SPDX-License-Identifier: MIT

def skip(ok)
  val = 0 # RuboCop is silent here
  if ok
    val = 5
    val += 5 if ok
    p val
  end
  0
end

def catch(ok)
  num = 0 # RuboCop sees this
  if ok
    num = 5
    num += 5
    p num
  end
  0
end
```

```bash
$ bundle exec rubocop useless.rb
```

```bash
Inspecting 1 file
W

Offenses:

useless.rb:17:3: W: [Correctable] Lint/UselessAssignment: Useless assignment to variable - num.
  num = 0 # RuboCop sees this
  ^^^

1 file inspected, 1 offense detected, 1 offense autocorrectable
```

Let’s find a solution so we can catch these cases consistently across the codebase.

kreinba:
Issue #485 reports that RuboCop's `Lint/UselessAssignment` misses a class of useless writes: a local variable is initialized just before an `if`/`unless`/`case` that overwrites it in every reachable branch without ever reading the initial value. The cop stays silent because, syntactically, the variable is still "read" on the false path when no `else` is present.

This pull request adds a minitest case under `test/` that parses every Ruby file in `lib/` with the `parser` gem (already in the bundle via RuboCop), walks each sequence of statements, and asserts no `lvasgn` matches the shape: cond does not read the variable, the truthy branch overwrites it without reading first, and either both branches overwrite or no later statement reads the initial value. The `lib/` tree is clean today, so the test passes; new regressions will fail it with a file-and-line list and a pointer back to the issue. I spot-checked the detector against the example in the issue body and it reports the expected offense.

Checks run: `bundle exec rake rubocop` (clean), `bundle exec rake yard` (clean), `bundle exec rake test` (529 tests, 1 failure — `TestLogged#test_proper_logging`, which fails the same way on `master` due to a string-truncation expectation and is unrelated to this change), and the new file in isolation under `PICKS=yes`.

Closes #485

### DIFF
--- a/test/test_no_useless_conditional_assignment.rb
+++ b/test/test_no_useless_conditional_assignment.rb
@@ -1,0 +1,94 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'parser/current'
+require_relative 'test__helper'
+
+# Detects an antipattern that RuboCop's +Lint/UselessAssignment+ misses:
+# a local variable is initialized just before an +if+/+unless+/+case+ that
+# overwrites the variable in every reachable branch, never having read the
+# initial value. See yegor256/factbase#485.
+class TestNoUselessConditionalAssignment < Factbase::Test
+  ROOT = File.expand_path('../lib', __dir__)
+
+  def test_lib_is_free_of_useless_conditional_assignments
+    offenses = []
+    Dir.glob(File.join(ROOT, '**', '*.rb')).each do |path|
+      ast =
+        begin
+          Parser::CurrentRuby.parse(File.read(path))
+        rescue Parser::SyntaxError
+          nil
+        end
+      next if ast.nil?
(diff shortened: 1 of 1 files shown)


0crat:
@kreinba **Excellent work!** 🎉 You've earned **+24 points** for this contribution (+24 base, +7 for 94 hits-of-code, -7 cap adjustment). [Keep the momentum going](https://www.yegor256.com/2018/03/06/speed-vs-quality.html)! Your total score is now [+1881](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #649, opened 2026-06-14
Restore eager materialization in IndexedQuery#each (#648)

### LINKED ISSUE #648: IndexedQuery#each no longer materializes results, so inserts during iteration change query matches
In `lib/factbase/indexed/indexed_query.rb:38-46`, `IndexedQuery#each` streams facts straight from `@origin` and yields them one by one. Until 0.19.12 it first drained the origin into an array (`a = @origin.each(fb, params).to_a`) and only then yielded each element, so the whole result set was computed before the caller saw the first fact.

Because the yields now interleave with the origin query's per-fact term evaluation, a caller that inserts facts inside the loop changes which later facts match. An `(empty ...)` sub-term used as a dedup guard, for example, starts seeing the facts that were just inserted. The eager `.to_a` previously made each iteration a stable snapshot. This pure-factbase snippet prints `3` on 0.19.11 and `1` on 0.19.12:

```ruby
require 'factbase'
fb = Factbase::IndexedFactbase.new(Factbase.new)
3.times { |i| fb.insert.then { |f| f.kind = 'in'; f.id = i } }
created = 0
fb.query("(and (eq kind 'in') (empty (eq kind 'out')))").each do |_f|
  fb.insert.then { |n| n.kind = 'out' }
  created += 1
end
puts created
```

This pattern (insert during iteration, guarded by `(empty ...)`) is what `Fbe.conclude` relies on, and it broke downstream. The fix is to restore eager materialization in `each`: collect `@origin.each(fb, params)` into an array, then yield from that array, so the iteration stays a consistent snapshot even when the caller mutates the factbase mid-loop.

yegor256:
Fixes #648.

## Problem

Since 0.19.12, `IndexedQuery#each` streamed facts straight from `@origin`, yielding them one by one. Because the yields now interleaved with the origin query's per-fact term evaluation, a caller that inserts facts inside the loop changed which later facts matched. An `(empty ...)` sub-term used as a dedup guard, for example, started seeing the facts that were just inserted — which broke `Fbe.conclude` downstream.

Before 0.19.12, `each` first drained the origin into an array (`a = @origin.each(fb, params).to_a`) and only then yielded each element, making each iteration a stable snapshot.

## Fix

Restore eager materialization: collect `@origin.each(fb, params)` into an array before yielding, so iteration stays a consistent snapshot even when the caller mutates the factbase mid-loop.

## Test

Added `test_materializes_before_iterating`, which reproduces the scenario from the issue (the snippet that prints `3` on 0.19.11 and `1` on 0.19.12). Verified it fails without the fix and passes with it.

https://claude.ai/code/session_01XdpmLXjrvfEVNt95rHcdDj

---
_Generated by [Claude Code](https://claude.ai/code/session_01XdpmLXjrvfEVNt95rHcdDj)_

### DIFF
--- a/lib/factbase/indexed/indexed_query.rb
+++ b/lib/factbase/indexed/indexed_query.rb
@@ -38,7 +38,7 @@ class Factbase::IndexedQuery
   def each(fb = @fb, params = {})
     return to_enum(__method__, fb, params) unless block_given?
     n = 0
-    @origin.each(fb, params) do |f|
+    @origin.each(fb, params).to_a.each do |f|
       yield(Factbase::IndexedFact.new(f, @idx, @fresh))
       n += 1
     end
--- a/test/factbase/indexed/test_indexed_query.rb
+++ b/test/factbase/indexed/test_indexed_query.rb
@@ -201,6 +201,21 @@ class TestIndexedQuery < Factbase::Test
     end
   end
 
+  def test_materializes_before_iterating
+    fb = Factbase::IndexedFactbase.new(Factbase.new)
+    3.times do |i|
+      f = fb.insert
+      f.kind = 'in'
+      f.id = i
+    end
+    created = 0
+    fb.query("(and (eq kind 'in') (empty (eq kind 'out')))").each do |_f|
+      fb.insert.then { |n| n.kind = 'out' }
+      created += 1
+    end
+    assert_equal(3, created)
+  end
+
   def test_deletes_too
     fb = Factbase::IndexedFactbase.new(Factbase.new)
     fb.insert.foo = 1


0crat:
@yegor256 According to our [work policy](https://www.yegor256.com/2014/04/15/github-guidelines.html), your branch name "`claude/laughing-dirac-recehd`" doesn't follow our naming convention, resulting in a **-6 point penalty** as specified in our guidelines. Branch names should match the ticket number you're solving - in this case, the correct name would be "`648`". Your running score is [+338](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); please check your [Zerocracy account](https://www.zerocracy.com/) and follow the policy to avoid future penalties.

0crat:
@yegor256 Thanks for the contribution! You've earned **+12 points** for this: **+24** as a basis; **-8** for the lack of code review (our policy requires peer review to maintain code quality); **-4** for contributing only **17** hits-of-code (under our **30** minimum threshold). Consider submitting larger, well-reviewed contributions to maximize your rewards per our [work policy](https://www.yegor256.com/2018/03/06/speed-vs-quality.html). Your running score is [+310](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 🚀

PR #664, opened 2026-07-06
#644: fix docstring in Factbase::Inv block signature

### LINKED ISSUE #644: lib/factbase/inv.rb: docstring on Factbase::Inv documents a block signature the class does not implement
The YARD docstring on `Factbase::Inv` in `lib/factbase/inv.rb` (lines 12 to 20) describes a block signature that the class does not implement. The comment shows `Factbase::Inv.new(Factbase.new) do |f, fbt|` and an example body of `assert !f['when'].nil?`, claiming the first block argument is "the fact just touched" and the second is the factbase.

The actual implementation at line 67 invokes the block as `@block.call(k[0..-2], args[1])`, passing the property name (a `String`) and the value being assigned. The test `test/factbase/test_inv.rb` confirms this: `Factbase::Inv.new(Factbase.new) do |p, v| ... end`. A user copying the docstring will get a `String` where they expect a `Factbase::Fact`, and `'when'['when']` raises `TypeError` at runtime rather than checking the invariant the snippet appears to assert.

Rewrite the example to match the real signature, something like `Factbase::Inv.new(Factbase.new) do |property, value| raise unless property != 'when' || !value.nil? end`, and adjust the prose so the two block parameters are documented as the property name and the value being set.

VasilevNStas:
Closes #644.

Per #644, the YARD docstring on `Factbase::Inv` documented the block signature as `|f, fbt|` (fact, factbase), but the actual implementation passes the property name and value. Updated the docstring to show `|p, v|` with correct descriptions, and added a PDD puzzle for a follow-up test that validates docstring examples against real API signatures.

### DIFF
--- a/lib/factbase/inv.rb
+++ b/lib/factbase/inv.rb
@@ -7,17 +7,25 @@ require 'decoor'
 require 'others'
 require_relative '../factbase'
 
+# @todo #644:30min Add a test that validates YARD docstring examples
+#  match the actual API signatures. The docstring in this class was
+#  recently fixed because the block signature was documented as
+#  |f, fbt| but the actual implementation passes |property, value|.
+#  A general test that parses docstrings and checks their examples
+#  against the real method signatures would prevent this class of
+#  bugs from recurring across the codebase.
+#
 # A decorator of a Factbase, that checks invariants on every set.
 #
-# For example, you can use this decorator if you want to check that every
-# fact has +when+:
+# For example, you can use this decorator if you want to make sure that
+# no property +b+ is set to a +String+ value:
 #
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 plz review this PR

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2044](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #666, opened 2026-07-06
#655: guard against nil in Env#evaluate

### LINKED ISSUE #655: `Factbase::Env` crashes with `NoMethodError` when first operand resolves to `nil`
**What happens:** In `lib/factbase/terms/env.rb` (line 25), `_values(0, fact, maps, fb)[0]` is called without checking whether `_values` returned `nil` first. When the first operand is a `Symbol` referring to an attribute that does not exist in the current fact, `_values` returns `nil`, and `nil[0]` raises `NoMethodError: undefined method '[]' for nil:NilClass`. The same issue exists on line 26 for the default-value operand inside the `ENV.fetch` block. All other conversion terms (`to_float`, `to_time`, `to_integer`, `arithmetic`) guard with `return if vv.nil?` before indexing.

**What should happen:** `(env :missing_attr 'fallback')` should return `'fallback'` when the attribute is absent in the current fact, consistent with how other terms handle missing symbol operands.

VasilevNStas:
Closes #655.

Added nil-guards before indexing into `_values` results in `Factbase::Env#evaluate`, matching the pattern used by all other conversion terms (`to_float`, `to_time`, `to_integer`, `to_string`).

Also added two tests:
- `test_missing_attr_returns_nil` — first operand is a sym for a missing attribute
- `test_missing_default_returns_nil` — both operands are missing attributes

### DIFF
--- a/lib/factbase/terms/env.rb
+++ b/lib/factbase/terms/env.rb
@@ -22,7 +22,12 @@ class Factbase::Env < Factbase::TermBase
   # @return [String] The value of the environment variable or the default
   def evaluate(fact, maps, fb)
     assert_args(2)
-    n = _values(0, fact, maps, fb)[0]
-    ENV.fetch(n.upcase) { _values(1, fact, maps, fb)[0] }
+    n = _values(0, fact, maps, fb)
+    return if n.nil?
+    ENV.fetch(n[0].upcase) do
+      d = _values(1, fact, maps, fb)
+      return if d.nil?
+      d[0]
+    end
   end
 end
--- a/test/factbase/terms/test_env.rb
+++ b/test/factbase/terms/test_env.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
-require_relative '../../../lib/factbase/term'
-require_relative '../../../lib/factbase/terms/env'
 # SPDX-FileCopyrightText: Copyright (c) 2024-2026 Yegor Bugayenko
 # SPDX-License-Identifier: MIT
+
+require_relative '../../../lib/factbase/term'
+require_relative '../../../lib/factbase/terms/env'
 
 require_relative '../../test__helper'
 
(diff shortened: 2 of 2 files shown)


VasilevNStas:
@yegor256 plz review this PR

lib/factbase/terms/env.rb
```diff
@@ -22,7 +22,12 @@ def initialize(operands)
   # @return [String] The value of the environment variable or the default
   def evaluate(fact, maps, fb)
     assert_args(2)
-    n = _values(0, fact, maps, fb)[0]
-    ENV.fetch(n.upcase) { _values(1, fact, maps, fb)[0] }
+    n = _values(0, fact, maps, fb)
+    return if n.nil?
+    ENV.fetch(n[0].upcase) do
```
morphqdd, line 27: `n[0].upcase` still assumes the resolved attribute is a String — a non-string operand (e.g. an integer property) would raise `NoMethodError` here. Pre-existing and out of scope for #655, just flagging the next rough edge.

morphqdd (approved):
Resolves #655 cleanly — both the attribute operand and the lazily-evaluated default are nil-guarded, and the two new tests cover both paths. LGTM. One adjacent note inline.

0crat:
@morphqdd Thanks for the review! You've earned +13 points for this: +18 as a basis; -5 for very few (1) comments.  Your running score is [+676](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2084](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #668, opened 2026-07-06
#660: fix nil error message in Check#it when expression is short

### LINKED ISSUE #660: `Check#it` in `Rules` shows `nil` instead of the rule expression in the error message when the expression is 32 characters or shorter
## What happens

`Check#it` in `lib/factbase/rules.rb` builds the error message for a fact that fails a rule like
this:

```ruby
def it(fact, fb)
  return if Factbase::Syntax.new(@expr).to_term.evaluate(fact, [], fb)
  e = "#{@expr[0..32]}..." if @expr.length > 32
  raise(ArgumentError, "The fact doesn't match the #{e.inspect} rule: #{fact}")
end
```

`e` is only assigned a value when `@expr.length > 32` — for any shorter rule expression, the `if`
guard is false, `e` stays `nil` (Ruby hoists the local variable but never assigns it), and
`e.inspect` in the message literally renders the four characters `nil`, not the rule that actually
failed.

Verified directly, with a typical short rule like `(exists foo)` (12 characters, well under the
32-character threshold):

```
expr.length=12
e=nil
message: The fact doesn't match the nil rule: <fact>
```

So a failing assertion against essentially any everyday rule — anything up to 32 characters, which
covers most simple `(exists ...)`, `(eq ...)`, `(gt ...)` checks — produces a message like
`"The fact doesn't match the nil rule: ..."`, which names no rule at all and actively misleads
whoever is debugging the failure into thinking something about `nil` is involved. The
32-character branch is presumably meant to *truncate* long expressions for a readable message, but
its `if` guard accidentally also suppresses the *untruncated* case entirely instead of falling
back to the full expression.

## What should happen

The message should always include the actual rule expression — truncated when it's long, shown in
full otherwise — for example:

```ruby
e = @expr.length > 32 ? "#{@expr[0..31]}..." : @expr
raise(ArgumentError, "The fact doesn't match the #{e.inspect} rule: #{fact}")
```

producing `"The fact doesn't match the \"(exists foo)\" rule: ..."` instead of the current
`"The fact doesn't match the nil rule: ..."`.

VasilevNStas:
Closes #660.

The `if` modifier at `lib/factbase/rules.rb:125` only assigned `e` when `@expr.length > 32`. For shorter expressions, `e` stayed `nil` and the error message showed `nil` instead of the actual rule.

Replaced with a ternary so `e` is always set. Also fixed the truncation boundary from `0..32` (33 chars) to `0..31` (32 chars) so the truncated form fits.

Added two tests:
- `test_error_message_includes_expression` — short rule shows in message
- `test_error_message_truncates_long_expression` — long rule is truncated with `...`

### DIFF
--- a/lib/factbase/rules.rb
+++ b/lib/factbase/rules.rb
@@ -122,8 +122,10 @@ class Factbase::Rules
 
     def it(fact, fb)
       return if Factbase::Syntax.new(@expr).to_term.evaluate(fact, [], fb)
-      e = "#{@expr[0..32]}..." if @expr.length > 32
-      raise(ArgumentError, "The fact doesn't match the #{e.inspect} rule: #{fact}")
+      raise(
+        ArgumentError,
+        "The fact doesn't match the #{(@expr.length > 32 ? "#{@expr[0..31]}..." : @expr).inspect} rule: #{fact}"
+      )
     end
   end
 
--- a/test/factbase/test_rules.rb
+++ b/test/factbase/test_rules.rb
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
 require_relative '../../lib/factbase'
 require_relative '../../lib/factbase/pre'
 require_relative '../../lib/factbase/rules'
-# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Yegor Bugayenko
-# SPDX-License-Identifier: MIT
 
 require_relative '../test__helper'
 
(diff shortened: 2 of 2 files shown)


VasilevNStas:
@yegor256 please review this PR

lib/factbase/rules.rb
```diff
@@ -122,8 +122,10 @@ def initialize(expr)
 
     def it(fact, fb)
       return if Factbase::Syntax.new(@expr).to_term.evaluate(fact, [], fb)
-      e = "#{@expr[0..32]}..." if @expr.length > 32
-      raise(ArgumentError, "The fact doesn't match the #{e.inspect} rule: #{fact}")
+      raise(
+        ArgumentError,
+        "The fact doesn't match the #{(@expr.length > 32 ? "#{@expr[0..31]}..." : @expr).inspect} rule: #{fact}"
```
morphqdd, line 127: Correct, but the nested ternary-inside-interpolation-inside-a-multiline-string is dense. Extracting a local first — `shown = @expr.length > 32 ? "#{@expr[0..31]}..." : @expr` — and interpolating `shown.inspect` would read more clearly.

morphqdd (approved):
Fixes #660 — the previous `e = "..." if @expr.length > 32` left `e` as nil for short expressions, so the message printed `"nil"`; the inline ternary now always renders the expression (truncated only when long). The two tests pin it well. One readability nit inline.

0crat:
@morphqdd Thanks for the review! You've earned +13 points for this: +18 as a basis; -5 for very few (1) comments.  Your running score is [+689](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2128](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #669, opened 2026-07-06
#661: rescue ArgumentError in to_float and to_integer like to_time

### LINKED ISSUE #661: `to_float` and `to_integer` terms do not rescue `ArgumentError` on invalid input, unlike `to_time`
## What happens

`Factbase::ToTime#evaluate` wraps its conversion in a rescue block and raises a descriptive
`RuntimeError`:

```ruby
def parse(value)
  Time.parse(value.to_s)
rescue ArgumentError => e
  raise(RuntimeError, "Cannot parse '#{value}' as Time in (to_time ...): #{e.message}")
end
```

`Factbase::ToFloat#evaluate` and `Factbase::ToInteger#evaluate` do not follow the same pattern —
they call `Float(vv[0])` / `Integer(vv[0])` directly, with no rescue at all:

```ruby
# ToFloat
def evaluate(fact, maps, fb)
  assert_args(1)
  vv = _values(0, fact, maps, fb)
  return if vv.nil?
  Float(vv[0])
end

# ToInteger
def evaluate(fact, maps, fb)
  assert_args(1)
  vv = _values(0, fact, maps, fb)
  return if vv.nil?
  Integer(vv[0])
end
```

So an invalid value such as `"abc"` raises a raw `ArgumentError` instead of the kind of contextual
error `to_time` produces. That `ArgumentError` then reaches `Term#evaluate`'s generic
`rescue StandardError` in `lib/factbase/term.rb`:

```ruby
rescue StandardError => e
  raise(RuntimeError, "#{e.message.inspect} at #{self} at #{e.backtrace[0]}")
end
```

which wraps the message with `.inspect`, double-escaping it. Verified directly:

```
message: invalid value for Float(): "abc"
inspect: "invalid value for Float(): \"abc\""
wrapped: "invalid value for Float(): \"abc\"" at <term> at <backtrace>
```

So a failing `(to_float ...)` or `(to_integer ...)` term produces a message like
`"\"invalid value for Float(): \\\"abc\\\"\" at (to_float f.something) at ..."` — a
backslash-escaped string inside a string, naming neither the term's own operand context nor the
offending value in a readable way, unlike `to_time`'s clean
`"Cannot parse 'abc' as Time in (to_time ...): invalid date"`.

## What should happen

`ToFloat#evaluate` and `ToInteger#evaluate` should rescue `ArgumentError` the same way
`ToTime#evaluate` does, raising a `RuntimeError` with a human-readable, term-specific message —
e.g. `"Cannot convert 'abc' to Float in (to_float ...): ..."` and
`"Cannot convert 'abc' to Integer in (to_integer ...): ..."` — instead of letting the raw
`ArgumentError` fall through to the generic, doubly-escaped wrapping in `Term#evaluate`.

VasilevNStas:
Closes #661.

Added rescue blocks in `ToFloat#evaluate` and `ToInteger#evaluate` matching the pattern used by `ToTime#parse`, so invalid input produces a clean `RuntimeError` instead of a raw `ArgumentError` that gets double-escaped by `Term#evaluate`.

Messages:
- `Cannot convert '<value>' to Float in (to_float ...): <reason>`
- `Cannot convert '<value>' to Integer in (to_integer ...): <reason>`

### DIFF
--- a/lib/factbase/terms/to_float.rb
+++ b/lib/factbase/terms/to_float.rb
@@ -29,8 +29,8 @@ class Factbase::ToFloat < Factbase::TermBase
   private
 
   def to_float(value)
-    Float(value)
-  rescue ArgumentError => e
+    Float(value.to_s)
+  rescue ArgumentError, TypeError => e
     raise(RuntimeError, "Cannot convert '#{value}' to Float in (to_float ...): #{e.message}")
   end
 end
--- a/lib/factbase/terms/to_float.rb
+++ b/lib/factbase/terms/to_float.rb
@@ -23,6 +23,14 @@ class Factbase::ToFloat < Factbase::TermBase
     assert_args(1)
     vv = _values(0, fact, maps, fb)
     return if vv.nil?
-    Float(vv[0])
+    to_float(vv[0])
+  end
+
+  private
+
+  def to_float(value)
+    Float(value)
+  rescue ArgumentError => e
+    raise(RuntimeError, "Cannot convert '#{value}' to Float in (to_float ...): #{e.message}")
   end
 end
--- a/lib/factbase/terms/to_integer.rb
+++ b/lib/factbase/terms/to_integer.rb
@@ -23,14 +23,14 @@ class Factbase::ToInteger < Factbase::TermBase
     assert_args(1)
     vv = _values(0, fact, maps, fb)
     return if vv.nil?
-    to_int(vv[0])
+    to_integer(vv[0])
   end
 
   private
 
-  def to_int(value)
-    Integer(value)
-  rescue ArgumentError => e
+  def to_integer(value)
+    Integer(value.to_s)
+  rescue ArgumentError, TypeError => e
     raise(RuntimeError, "Cannot convert '#{value}' to Integer in (to_integer ...): #{e.message}")
   end
 end
--- a/lib/factbase/terms/to_integer.rb
+++ b/lib/factbase/terms/to_integer.rb
@@ -23,6 +23,14 @@ class Factbase::ToInteger < Factbase::TermBase
     assert_args(1)
     vv = _values(0, fact, maps, fb)
     return if vv.nil?
-    Integer(vv[0])
+    to_int(vv[0])
+  end
+
+  private
+
+  def to_int(value)
+    Integer(value)
+  rescue ArgumentError => e
+    raise(RuntimeError, "Cannot convert '#{value}' to Integer in (to_integer ...): #{e.message}")
   end
 end
--- a/test/factbase/terms/test_to_float.rb
+++ b/test/factbase/terms/test_to_float.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
-require_relative '../../../lib/factbase/term'
-require_relative '../../../lib/factbase/terms/to_float'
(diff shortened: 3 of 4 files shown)


VasilevNStas:
@yegor256 plz review

morphqdd (commented):
Thanks for taking #661 — this is the right shape: extract a small helper, rescue `ArgumentError`, and re-raise a clean positioned `RuntimeError` mirroring `to_time`, with tests for both terms. It resolves the case I reported (a raw `ArgumentError` from `Float('abc')`/`Integer('abc')` double-escaping through `Term#evaluate`).

Two things worth tightening before merge, inline — the first is the same class of leak for a neighbouring input type.

lib/factbase/terms/to_float.rb
```diff
@@ -23,6 +23,14 @@ def evaluate(fact, maps, fb)
...
     vv = _values(0, fact, maps, fb)
     return if vv.nil?
-    Float(vv[0])
+    to_float(vv[0])
+  end
+
+  private
+
+  def to_float(value)
+    Float(value)
```
morphqdd: `Float(value)` only guards `ArgumentError`, but `Float`/`Integer` raise `TypeError` (not `ArgumentError`) for non-string, non-numeric operands — e.g. `Float(nil)` or `Float([])`. So `(to_float ...)` on a nil/array-valued property still leaks a raw `TypeError`, which is the same double-escape problem #661 is about. Note `to_time` sidesteps this entirely by coercing first: `Time.parse(vv[0].to_s)`. Consider either rescuing `TypeError` alongside `ArgumentError`, or matching `to_time` and calling `Float(value.to_s)` — otherwise the fix is only half of the leak.

lib/factbase/terms/to_integer.rb
```diff
@@ -23,6 +23,14 @@ def evaluate(fact, maps, fb)
     assert_args(1)
     vv = _values(0, fact, maps, fb)
     return if vv.nil?
-    Integer(vv[0])
+    to_int(vv[0])
+  end
+
+  private
+
+  def to_int(value)
```
morphqdd: Minor naming inconsistency: this helper is `to_int` while its `to_float` counterpart is `to_float` (full word). Same `TypeError` note as on `to_float` applies here too.

VasilevNStas:
@morphqdd good points. Fixed in 48351cc:
- Added `.to_s` coercion before `Float()`/`Integer()` (like `to_time` does with `value.to_s`)
- Also rescue `TypeError` alongside `ArgumentError` as defensive measure
- Renamed `to_int` to `to_integer` for consistency

0crat:
@vasilevnstas Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2152](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the review! You've earned +13 points for this: +18 as a basis; -5 for very few (2) comments.  Your running score is [+795](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #673, opened 2026-07-08
#672: fix to_integer crash on Float values

### LINKED ISSUE #672: `to_integer` crashes on Float values because `Integer(value.to_s)` rejects decimal strings
In `lib/factbase/terms/to_integer.rb:32`, `ToInteger#to_integer` converts via `Integer(value.to_s)`. When the operand is a `Float`, `value.to_s` yields a decimal string like `"5961600.0"`, and `Integer("5961600.0")` raises `ArgumentError: invalid value for Integer(): "5961600.0"`. Before this line was introduced (0.19.13 and earlier used `Integer(vv[0])`), `Integer(5961600.0)` truncated fine, so this is a regression in 0.19.14.

The break comes from the fix for #661, which copied `to_time`'s `Time.parse(value.to_s)` shape into `to_integer`. But `Integer()` already accepts numeric arguments directly, so the extra `.to_s` is not only unnecessary, it converts Floats into decimal strings that `Integer()` refuses to parse. This breaks the common `(to_integer (minus when X))` idiom, since subtracting two times produces a Float number of seconds. (`to_float` is unaffected because `Float()` does parse decimal strings.)

The fix is to drop `.to_s` on line 32, calling `Integer(value)` instead of `Integer(value.to_s)`. The surrounding `rescue ArgumentError, TypeError` from #661 stays and still catches genuinely invalid input like `Integer("abc")`, so the descriptive-error behavior is preserved while Float truncation works again.

yegor256:
Closes #672.

## Problem

`ToInteger#to_integer` converted values via `Integer(value.to_s)`. When the operand is a `Float`, `value.to_s` yields a decimal string like `"5961600.0"`, and `Integer("5961600.0")` raises `ArgumentError: invalid value for Integer()`. This was a regression introduced in 0.19.14 by the fix for #661 (which copied `to_time`'s `Time.parse(value.to_s)` shape). It broke the common `(to_integer (minus when X))` idiom, since subtracting two times produces a Float number of seconds.

## Fix

Drop the `.to_s` and call `Integer(value)` directly. `Integer()` already accepts numeric arguments and truncates Floats, so `Integer(5961600.0)` returns `5961600`. The surrounding `rescue ArgumentError, TypeError` from #661 stays and still catches genuinely invalid input like `Integer("abc")`, preserving the descriptive-error behavior.

## Tests

Added `test_truncates_float` covering the Float truncation path. All `to_integer` tests pass and RuboCop is clean.

https://claude.ai/code/session_01FhB2Mvved23UGNgGcE1UWX

---
_Generated by [Claude Code](https://claude.ai/code/session_01FhB2Mvved23UGNgGcE1UWX)_

### DIFF
--- a/lib/factbase/terms/to_integer.rb
+++ b/lib/factbase/terms/to_integer.rb
@@ -29,7 +29,7 @@ class Factbase::ToInteger < Factbase::TermBase
   private
 
   def to_integer(value)
-    Integer(value.to_s)
+    Integer(value)
   rescue ArgumentError, TypeError => e
     raise(RuntimeError, "Cannot convert '#{value}' to Integer in (to_integer ...): #{e.message}")
   end
--- a/test/factbase/terms/test_to_integer.rb
+++ b/test/factbase/terms/test_to_integer.rb
@@ -17,6 +17,10 @@ class TestToInteger < Factbase::Test
     assert_equal('Integer', Factbase::ToInteger.new([[42, 'hello']]).evaluate(fact, [], Factbase.new).class.to_s)
   end
 
+  def test_truncates_float
+    assert_equal(5_961_600, Factbase::ToInteger.new([[5_961_600.0]]).evaluate(fact, [], Factbase.new))
+  end
+
   def test_rejects_invalid_value
     t = Factbase::ToInteger.new(['abc'])
     assert_includes(


0crat:
@yegor256 Thanks for the contribution! You've earned +12 points for this: +24 as a basis; -8 for the lack of code review; -4 for too few (6) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+491](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #678, opened 2026-07-11
#677: exclude each consumers from query timeout

### LINKED ISSUE #677: Impatient#each times the consumer block again, regressing #264
`Factbase::Impatient::Query#each` (`lib/factbase/impatient.rb:58`) again runs the caller's block inside the timeout: the `yield(f)` on line 63 sits within the `impatient('each')` region, so any per-fact work the consumer does (network calls, etc.) is charged against the query timeout. This is exactly the problem fixed under #264 ("only fail on long queries, not blocks inside each"); it regressed in commit fb0a038 ("Performance improvements: lazy iteration..."), which replaced the materialize-then-yield form from 7373ad8 with a streaming loop that puts `yield` back inside the timed block.

It is wrong because the timeout is meant to bound query execution, not the consumer's processing. In zerocracy/judges-action (fbe wraps the factbase with `timeout: 60`), judges make a GitHub API call per fact inside the `each` block, so a legitimate loop dies with a misleading `each() timed out after 60s ... fb size is 17923` that blames the factbase size. Both timed-out queries from that run execute in 0.06s and 0.35s on a synthetic 17,923-fact base through the full decorator stack (Impatient->Logged->Sync->Indexed->Cached); the 60s is entirely GitHub latency. Same symptom as #321.

Fix: restore the form from 7373ad8 — wrap only fact production in `impatient('each')` (`... .each(fb, params).to_a`) and iterate/`yield` the resulting array outside the timeout. Runaway queries are still caught, because `empty`/`unique`/`join` terms evaluate during `.to_a`.

Failing run: https://github.com/objectionary/eo/actions/runs/29140067673/job/86511372525

w3lld1:
## Summary

- materialize `Impatient#each` query results inside the configured timeout
- invoke the caller's consumer block after the timed query has completed
- add regression coverage showing slow consumer work is not charged to the query timeout

## Validation

- `git diff --check`
- The focused Ruby test suite could not be run locally because Ruby is not installed on this runner, and its Docker daemon is not accessible. The upstream Ruby 3.4 CI should exercise the regression test on Linux, macOS, and Windows.

Fixes #677

### DIFF
--- a/lib/factbase/impatient.rb
+++ b/lib/factbase/impatient.rb
@@ -57,9 +57,10 @@ class Factbase::Impatient
 
     def each(fb = @fb, params = {}, &)
       return to_enum(__method__, fb, params) unless block_given?
-      facts = impatient('each') do
-        @fb.query(@term, @maps).each(fb, params).to_a
-      end
+      facts =
+        impatient('each') do
+          @fb.query(@term, @maps).each(fb, params).to_a
+        end
       n = 0
       facts.each do |f|
         yield(f)
--- a/lib/factbase/impatient.rb
+++ b/lib/factbase/impatient.rb
@@ -57,12 +57,13 @@ class Factbase::Impatient
 
     def each(fb = @fb, params = {}, &)
       return to_enum(__method__, fb, params) unless block_given?
+      facts = impatient('each') do
+        @fb.query(@term, @maps).each(fb, params).to_a
+      end
       n = 0
-      impatient('each') do
-        @fb.query(@term, @maps).each(fb, params) do |f|
-          yield(f)
-          n += 1
-        end
+      facts.each do |f|
+        yield(f)
+        n += 1
       end
       n
     end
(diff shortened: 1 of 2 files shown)


w3lld1:
Fixed the RuboCop `Layout/MultilineAssignmentLayout` failure by moving the multiline assignment RHS below `=` and pushed commit `5894407b`. `git diff --check` passes. I could not rerun Ruby/RuboCop locally because Ruby is unavailable on this runner; the previous CI run completed all 565 tests successfully before reaching this style offense.

0crat:
@w3lld1 Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+20](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #680, opened 2026-07-13
#670: raise clean ArgumentError on zero divisor in div term

### LINKED ISSUE #670: `(div ...)` term crashes with raw `ZeroDivisionError` when the divisor is zero
## What happens

A query that uses the `(div a b)` term where the second operand evaluates to the integer `0` crashes with a raw `ZeroDivisionError` that leaks internal implementation details to the caller.

`Factbase::Arithmetic#evaluate` (used by `Factbase::Div`) finishes with:

```ruby
# lib/factbase/terms/arithmetic.rb:53
v.send(@op, r)
```

For `div`, `@op` is `:/`. When `r` is an integer `0`, `Integer#/` raises `ZeroDivisionError`, which the term does not handle. `Factbase::Term#evaluate` then re-wraps it, embedding the raw `"divided by 0"` text together with internal file paths and line numbers.

Reproduced locally against the current `master`:

```ruby
fb = Factbase.new
f = fb.insert
f.a = 10
f.b = 0
fb.query('(eq 1 (div a b))').each.to_a
```

Result:

```
RuntimeError: "\"divided by 0\" at (div a b) at .../lib/factbase/terms/arithmetic.rb:53:in 'Integer#/'"
  at (eq 1 (div a b)) at .../lib/factbase/term.rb:212:in 'Factbase::Term#evaluate'
```

A non-zero divisor works correctly (`(div 8 2)` matches as expected). Note also the inconsistency: a `Float` zero divisor (`0.0`) yields `Infinity` with no error, while an `Integer` `0` crashes — the same query can succeed or blow up depending only on the numeric type of the divisor.

This is reachable in real pipelines: any `(div ...)` computing a ratio or average (success rates, per-item averages, etc.) hits it whenever the denominator is legitimately `0` — e.g. before any data has accumulated — and takes down the judge with a stack-leaking error rather than a clean failure.

## What should happen

`Factbase::Arithmetic` should detect a zero divisor for the `div` operation and raise a clean, domain-level error (for example `Cannot divide by zero in (div ...)`), consistent with how other terms report bad input, instead of letting a raw `ZeroDivisionError` with internal paths and line numbers escape through `Term#evaluate`. The `Integer` and `Float` zero-divisor paths should also behave consistently.

morphqdd:
Closes #670.

`Factbase::Arithmetic#evaluate` let `Integer#/` raise a raw `ZeroDivisionError` out of `(div a b)` whenever the divisor evaluated to integer `0`, leaking internal file paths and line numbers through `Term#evaluate`'s re-wrap; a `Float` zero divisor meanwhile silently produced `Infinity`. Both zero-divisor paths now raise the same clean, domain-level `ArgumentError` (`Cannot divide by zero`), consistent with how the term already reports other bad input (wrong arity, unknown time units).

Two new tests cover the integer-zero and float-zero divisors; both fail on master (raw `ZeroDivisionError` / `Infinity` respectively).

Tests: test_div, test_math, test_query all green (43 tests, 0 failures); rubocop: no offenses.

### DIFF
--- a/lib/factbase/terms/arithmetic.rb
+++ b/lib/factbase/terms/arithmetic.rb
@@ -50,6 +50,7 @@ class Factbase::Arithmetic < Factbase::TermBase
           raise(ArgumentError, "Unknown time unit '#{units}' in '#{r}'")
         end
     end
+    raise(ArgumentError, 'Cannot divide by zero') if @op == :/ && r.is_a?(Numeric) && r.zero?
     v.__send__(@op, r)
   end
 end
--- a/test/factbase/terms/test_div.rb
+++ b/test/factbase/terms/test_div.rb
@@ -16,6 +16,26 @@ class TestDiv < Factbase::Test
     assert_equal(420, Factbase::Div.new([:balance, 100]).evaluate(fact('balance' => 42_000), [], Factbase.new))
   end
 
+  def test_div_by_integer_zero
+    t = Factbase::Div.new([:balance, 0])
+    assert_includes(
+      assert_raises(ArgumentError) do
+        t.evaluate(fact('balance' => 42), [], Factbase.new)
+      end.message,
+      'Cannot divide by zero'
+    )
+  end
+
+  def test_div_by_float_zero
+    t = Factbase::Div.new([:balance, 0.0])
+    assert_includes(
+      assert_raises(ArgumentError) do
+        t.evaluate(fact('balance' => 42.0), [], Factbase.new)
+      end.message,
+      'Cannot divide by zero'
+    )
+  end
+
   def test_div_times_not_supported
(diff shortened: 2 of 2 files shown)


morphqdd:

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+972](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #687, opened 2026-07-24
#686: treat empty env var as unset in Factbase::Env

### LINKED ISSUE #686: `Factbase::Env#evaluate` (`env` DSL term) doesn't fall back to default when the variable is set but empty
## What Happens

`Factbase::Env#evaluate` (`lib/factbase/terms/env.rb:27`, the `env` DSL term) does `ENV.fetch(n[0].upcase) { ... default ... }`. This only falls back to the default when the environment variable is unset. If the variable is set but empty (`''`), `fetch` returns the empty string instead of the default.

This breaks the common `(to_time (env 'TODAY' '#{Time.now.utc.iso8601}'))` pattern used by several downstream judges (for example in `zerocracy/judges-action`). If `TODAY` is set but empty, `Factbase::ToTime#parse` calls `Time.parse('')`, which raises `ArgumentError`, re-raised as `RuntimeError: Cannot parse '' as Time in (to_time ...)`.

Verified directly: setting `ENV['TODAY'] = ''` and evaluating `(to_time (env 'TODAY' '2026-01-01T00:00:00Z'))` against the real `factbase` gem raises exactly this unhandled `RuntimeError`, instead of falling back to the provided default.

Filed originally against `zerocracy/judges-action` as issue #1858, where @yegor256 pointed out the actual bug is in this gem, not the caller.

## What Should Happen

`Factbase::Env#evaluate` should treat an empty string the same as an unset variable and fall back to the default, for example:
```ruby
value = ENV[n.upcase]
if value.nil? || value.empty?
  default
else
  value
end
```

morphqdd:
Closes #686

`Factbase::Env#evaluate` (the `env` DSL term) used `ENV.fetch(n[0].upcase) { default }`, which only falls back to the default when the environment variable is unset. If the variable is set but empty (`''`), `fetch` returned the empty string instead of the default.

This breaks the common `(to_time (env 'TODAY' '#{Time.now.utc.iso8601}'))` pattern used by several downstream judges in zerocracy/judges-action. If `TODAY` is set but empty, `Factbase::ToTime#parse` calls `Time.parse('')`, which raises.

Fix: read the value with `ENV.fetch(n[0].upcase, nil)` and treat `nil` or an empty string the same way, falling back to the default in both cases.

Added a test for the empty-string case (`test_default_when_value_is_empty`).

Verified:
- All 6 tests in `test/factbase/terms/test_env.rb` pass (5 existing + 1 new).
- rubocop clean.

### DIFF
--- a/lib/factbase/terms/env.rb
+++ b/lib/factbase/terms/env.rb
@@ -24,10 +24,13 @@ class Factbase::Env < Factbase::TermBase
     assert_args(2)
     n = _values(0, fact, maps, fb)
     return if n.nil?
-    ENV.fetch(n[0].upcase) do
+    value = ENV.fetch(n[0].upcase, nil)
+    if value.nil? || value.empty?
       d = _values(1, fact, maps, fb)
       return if d.nil?
       d[0]
+    else
+      value
     end
   end
 end
--- a/test/factbase/terms/test_env.rb
+++ b/test/factbase/terms/test_env.rb
@@ -23,6 +23,11 @@ class TestEnv < Factbase::Test
     assert_equal('мой друг', Factbase::Env.new(['FOO', 'мой друг']).evaluate(fact, [], Factbase.new))
   end
 
+  def test_default_when_value_is_empty
+    ENV.store('FOO', '')
+    assert_equal('мой друг', Factbase::Env.new(['FOO', 'мой друг']).evaluate(fact, [], Factbase.new))
+  end
+
   def test_when_default_is_absent
     ENV.delete('FOO')
     t = Factbase::Env.new(['FOO'])


morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1036](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #618, opened 2026-06-10
Flatten#it sort_by crashes on incompatible value types


VasilevNStas:
Flatten#it uses sort_by { |m| m[@sorter] || [] }. If maps have values of incompatible types (e.g. Integer vs String) for the same key, Array#<=> raises ArgumentError.

VasilevNStas:
Fixed in PR #

VasilevNStas:
Fixed in #611

PR #705, opened 2026-09-01
#695: sort the printed facts by kind, then by value

### LINKED ISSUE #695: `Factbase::Flatten#it` sorts facts by `_id` as strings, so `to_json`/`to_yaml`/`to_xml` print fact 10 before fact 2
## What happens

`Factbase::Flatten#it` converts the sorter values to strings before sorting, so
integer `_id` values are ordered lexicographically instead of numerically:

https://github.com/yegor256/factbase/blob/master/lib/factbase/flatten.rb#L22-L28

```ruby
  def it
    @maps
      .sort_by { |m| Array(m[@sorter]).map(&:to_s) }
      .map { |m| m.sort.to_h }
      .map! { |m| m.transform_values { |v| v.size == 1 ? v[0] : v } }
  end
```

`Factbase::ToJSON`, `Factbase::ToYAML` and `Factbase::ToXML` all print through
this class, so as soon as a factbase holds ten or more facts, the printed order
is `1, 10, 11, 12, 2, 3, ...`. Sorting the printed facts by `_id` is the whole
point of this line (#61).

The `.map(&:to_s)` was introduced by 5fd4f461 as the fix for #618
(`sort_by` crashing on incompatible value types). Before that commit the line
was `.sort_by { |m| m[@sorter] || [] }` and the same program below printed
`1, 2, 3, ... 12`. This report is about the ordering, not about the crash
that #618 covered.

## Steps to reproduce

```ruby
require 'factbase'
require 'factbase/to_json'
fb = Factbase.new
(1..12).each { |i| fb.insert._id = i }
puts Factbase::ToJSON.new(fb).json
```

## Observed

```
[{"_id":1},{"_id":10},{"_id":11},{"_id":12},{"_id":2},{"_id":3},{"_id":4},{"_id":5},{"_id":6},{"_id":7},{"_id":8},{"_id":9}]
```

`Factbase::ToXML` and `Factbase::ToYAML` emit the same order, since they use the
same `Flatten`.

Reverting only line 24 to `.sort_by { |m| m[@sorter] || [] }` on the same
commit makes the very same script print:

```
[{"_id":1},{"_id":2},{"_id":3},{"_id":4},{"_id":5},{"_id":6},{"_id":7},{"_id":8},{"_id":9},{"_id":10},{"_id":11},{"_id":12}]
```

## Expected behavior

Facts printed by `ToJSON`, `ToYAML` and `ToXML` should be ordered by the natural
order of the sorter property, so a factbase with integer `_id` values from 1 to
12 prints them as 1, 2, 3, ..., 12, while still not raising when the sorter
values of different facts have incompatible types.

morphqdd:
`Flatten#it` turned every sorter value into a string before sorting, so integer
`_id`s were ordered by their digits and a factbase of twelve facts printed as
1, 10, 11, 12, 2, 3, ... through `ToJSON`, `ToYAML` and `ToXML` alike.

The `to_s` was there to stop `sort_by` raising when two facts hold values of
different types (#618). The key now carries the kind first and the value
second, so values of different kinds never reach a comparison with each other,
and numbers keep their own order.

`Flatten.new((1..12).map { |n| {'_id'=>[n]} }, '_id')` gives 1..12 with this
and 1, 10, 11, 12, 2, ... without it. The existing incompatible-types test
still passes, and four new ones cover numbers, strings, times and a mix.

Closes #695

### DIFF
--- a/lib/factbase/flatten.rb
+++ b/lib/factbase/flatten.rb
@@ -21,8 +21,26 @@ class Factbase::Flatten
   # @return [Array<HashMap>] The hashmaps, but improved
   def it
     @maps
-      .sort_by { |m| Array(m[@sorter]).map(&:to_s) }
+      .sort_by { |m| Array(m[@sorter]).map { |v| Factbase::Flatten.key(v) } }
       .map { |m| m.sort.to_h }
       .map! { |m| m.transform_values { |v| v.size == 1 ? v[0] : v } }
+  end
+
+  # The sorting key of one value: the name of its kind, then the value
+  # itself for the kinds that compare, and its text for the rest. Values
+  # of different kinds never meet in a comparison this way, which is what
+  # made the plain value raise (#618), while numbers stay in their own
+  # order instead of the order of their digits (#695).
+  # @param [Object] value The value of the sorter property
+  # @return [Array] The key to sort by
+  def self.key(value)
+    case value
+    when Numeric
+      [0, value, '']
+    when Time
+      [1, Float(value), '']
+    else
+      [2, 0, value.to_s]
+    end
   end
 end
--- a/test/factbase/test_flatten.rb
+++ b/test/factbase/test_flatten.rb
@@ -61,11 +61,6 @@ class TestFlatten < Factbase::Test
   end
 
   def test_sorts_values_of_mixed_kinds
-    assert_equal(
-      3,
-      Factbase::Flatten.new(
-        [{ 'x' => ['b'] }, { 'x' => [1] }, { 'x' => [Time.now] }], 'x'
-      ).it.size
-    )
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks for the fix — merged. Keying the sort by kind first and value second is a nice minimal fix: it keeps #618's incompatible-types safety while restoring natural numeric/chronological ordering, and the new tests cover numbers, strings, times, and mixed kinds.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1190](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #706, opened 2026-09-01
#694: delete what the query yielded, without asking it twice

### LINKED ISSUE #694: `Factbase::Query#delete!` re-evaluates the term on a raw `Fact`, so `(as ...)` in the query writes properties into facts it does not delete
## What happens

`Factbase::Query#delete!` evaluates the term a second time, and that second pass
wraps every map in a bare `Factbase::Fact` instead of the read-only
`Factbase::Accum` that `each()` uses:

https://github.com/yegor256/factbase/blob/master/lib/factbase/query.rb#L94-L104

```ruby
    @maps.delete_if do |m|
      d =
        if ids
          i = m[id]&.first
          i && ids.include?(i)
        else
          @term.evaluate(Factbase::Fact.new(m), @maps, fb)
        end
      deleted += 1 if d
      d
    end
```

In `each()` the fact is wrapped as `Factbase::Accum.new(f, extras, false)`
(line 55), so terms with side effects — `as`, `join` — write into the throw-away
`extras` hash. In `delete!` there is no such wrapper, so `(as p v)` calls
`fact.p = v` directly on the stored map. Every fact the term is evaluated
against gets the new property permanently, including the facts that do not match
and therefore survive the deletion.

The second pass runs whenever the matched facts have no `id` property (the
default is `_id`, which `Factbase#insert` does not assign), which is the common
case.

## Steps to reproduce

```ruby
require 'factbase'
fb = Factbase.new
%w[a b].each { |v| fb.insert.foo = v }
p fb.each.to_a
p fb.query("(and (as marked 42) (eq foo 'a'))").delete!
p fb.each.to_a
```

## Observed

```
[{"foo"=>["a"]}, {"foo"=>["b"]}]
1
[{"foo"=>["b"], "marked"=>[42]}]
```

The surviving fact `foo=b` never matched the query, yet it now carries
`marked=42`. Running the identical query through `each()` instead of `delete!`
leaves the factbase untouched.

The same double evaluation also breaks stateful terms. With three facts
`a`, `a`, `b`, the query `(unique foo)` matches two of them
(the first `a` and `b`), but `fb.query('(unique foo)').delete!` returns `1` and
leaves `["a", "a"]` behind, because the `unique` term keeps the values it saw
during the first pass.

## Expected behavior

`delete!` should delete exactly the facts that `each()` yields for the same
query, and it must not modify any fact in the factbase — a fact that is not
deleted has to come out of `delete!` byte-for-byte identical to how it went in.

morphqdd:
`delete!` ran the term a second time over bare `Factbase::Fact` wrappers, so a
query with `as` wrote into facts it did not delete, and a term that remembers,
like `unique`, answered differently the second time and deleted fewer facts
than `each` had yielded.

It now collects the facts of the single `each` pass and deletes those. A fact
tells which map it stands for through `origin`, so two facts of equal content
are still two facts, and the collection compares by identity rather than by
value.

The `id:` keyword is gone: it existed to match the two passes by `_id`, and
with one pass there is nothing to match. Nothing in the repository passed it.

Three tests cover it: the `as` case from the issue, the `unique` case, and two
facts of identical content.

Closes #694

### DIFF
--- a/lib/factbase/accum.rb
+++ b/lib/factbase/accum.rb
@@ -30,12 +30,6 @@ class Factbase::Accum
     @fact.all_properties | @props.keys
   end
 
-  # The map the decorated fact stands for.
-  # @return [Hash] That very map
-  def origin
-    @fact.origin
-  end
-
   others do |*args|
     k = args[0].to_s
     if k.end_with?('=')
--- a/lib/factbase/accum.rb
+++ b/lib/factbase/accum.rb
@@ -30,6 +30,12 @@ class Factbase::Accum
     @fact.all_properties | @props.keys
   end
 
+  # The map the decorated fact stands for.
+  # @return [Hash] That very map
+  def origin
+    @fact.origin
+  end
+
   others do |*args|
     k = args[0].to_s
     if k.end_with?('=')
--- a/lib/factbase/fact.rb
+++ b/lib/factbase/fact.rb
@@ -46,6 +46,19 @@ class Factbase::Fact
     @map.keys
   end
 
+  # The map this fact stands for, told apart from every other map.
+  #
+  # A fact is a view of one map of the factbase, and two facts of equal
+  # content are still two different facts. This tells them apart, so that
+  # a caller who has the fact can find the very map it came from, which is
+  # what `Query#delete!` needs in order to delete what it yielded rather
+  # than to run the query a second time (#694).
+  #
+  # @return [Hash] That very map
+  def origin
+    @map
+  end
+
   others do |*args|
     k = args[0].to_s
     if k.end_with?('=')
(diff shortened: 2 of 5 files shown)


yegor256:
Thanks for the fix — merged. Running `delete!` through the same `Accum`-wrapped single pass that `each` uses is exactly right: it stops `as` from writing into survivors and stops `unique` from being asked the same question twice, and the three new tests cover both failure modes plus the equal-content case.

---

morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1214](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #722, opened 2026-09-02
#701: refuse a negative position

### LINKED ISSUE #701: `Factbase::Nth` answers the last map for a negative position instead of refusing it
## What happens

`Factbase::Nth#evaluate` checks that the position is an Integer, but not that
it is a position at all:

```ruby
pos = @operands[0]
raise "An integer is expected, but #{pos} provided" unless pos.is_a?(Integer)
...
m = maps[pos]
return nil if m.nil?
```

Ruby reads a negative index from the end of the array, so a negative position
answers a fact instead of being refused:

```ruby
Factbase::Term.new(:nth, [-1, :foo]).evaluate(fact, maps, fb)  # => [4]
```

with `maps` holding one fact whose `foo` is 4. A position past the end answers
`nil`, which is the documented "no such map", so the two ends of the same
mistake behave differently: one is reported, the other is answered as if it
were a real position.

## What should happen

`(nth -1 f)` should be refused the way a non-integer position is, since the
term is documented as taking the n-th map and there is no minus-first map.

morphqdd:
`Nth#evaluate` checked that the position is an Integer but not that it is a position. Ruby reads a negative index from the end, so `(nth -1 f)` answered the last fact instead of being refused, while a position past the end answered nil. Two ends of the same mistake behaved differently.

Closes #701

### DIFF
--- a/lib/factbase/terms/nth.rb
+++ b/lib/factbase/terms/nth.rb
@@ -24,6 +24,7 @@ class Factbase::Nth < Factbase::TermBase
     assert_args(2)
     pos = @operands[0]
     raise(ArgumentError, "An integer is expected, but #{pos} provided") unless pos.is_a?(Integer)
+    raise(ArgumentError, "A non-negative position is expected, but #{pos} provided") if pos.negative?
     k = @operands[1]
     raise(ArgumentError, "A symbol is expected, but #{k} provided") unless k.is_a?(Symbol)
     m = maps[pos]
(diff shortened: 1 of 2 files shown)


morphqdd:

yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2382](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #724, opened 2026-09-02
#688: name the format operand the fact does not have

### LINKED ISSUE #688: `Factbase::Sprintf#evaluate` crashes with misleading error when format-string operand is a missing property
## What happens

`Factbase::Sprintf#evaluate` in `lib/factbase/terms/sprintf.rb` reads its first operand (the format string) without a nil guard:

```ruby
def evaluate(fact, maps, fb)
  formatted(
    _values(0, fact, maps, fb)[0],
    (1..(@operands.length - 1)).map { |i| _values(i, fact, maps, fb)&.first }
  )
end
```

Every other operand in the same method safely uses `&.first`, but the format string operand indexes into the result of `_values(0, ...)` directly with `[0]`.

`_values` (in `lib/factbase/terms/base.rb`) returns bare `nil` when the referenced property is absent from the current fact. When the format-string operand is a symbol referencing a missing property, `_values(0, ...)` returns `nil`, and `nil[0]` raises `NoMethodError: undefined method '[]' for nil`.

When `sprintf` is nested inside a real query, `Factbase::Term`'s dispatcher wraps this in a misleading message, since it can't tell the difference between "no such term" and an internal `NoMethodError`:

```
RuntimeError: Probably the term 'sprintf' is not defined at (sprintf $absent 'hi'): undefined method '[]' for nil
```

`sprintf` clearly is defined. The message hides the real cause, which is a missing property.

Reproduced directly:

```ruby
Factbase::Sprintf.new([:missing_prop, "x"]).evaluate(fact, maps, fb)
# => NoMethodError: undefined method '[]' for nil
```

And through the public query DSL:

```ruby
fb.query("(eq (sprintf $absent 'hi') 'hi')").each(&:to_s)
# => RuntimeError: Probably the term 'sprintf' is not defined at (sprintf $absent 'hi'): undefined method '[]' for nil
```

This is the same bug class as issue #655 (`Factbase::Env#evaluate`), already fixed: an unguarded `_values(0, ...)[0]` where every sibling call in the same method uses the safe `&.first`.

## What should happen

`(sprintf $missing_prop 'hi')` should behave consistently with the other operands: either treat a missing format-string property as `nil`/`false` (matching how the other operands are handled), or raise a clear, contextual error that names the term and the missing property, not an opaque `NoMethodError` misreported as "term not defined."

morphqdd:
Every operand but the first used `&.first`; the format string was indexed with `[0]` straight on the result of `_values`, which is nil when the referenced property is absent. `nil[0]` then raised `NoMethodError`, and the term dispatcher rewrote it as "Probably the term 'sprintf' is not defined", which hid the real cause.

The format operand is now read the same way as the others and, when it is not there, the error names it.

Closes #688

### DIFF
--- a/lib/factbase/terms/sprintf.rb
+++ b/lib/factbase/terms/sprintf.rb
@@ -22,8 +22,12 @@ class Factbase::Sprintf < Factbase::TermBase
   # @param [Factbase] fb Factbase to use for sub-queries
   # @return [String] The formatted string
   def evaluate(fact, maps, fb)
+    fmt = _values(0, fact, maps, fb)&.first
+    if fmt.nil?
+      raise(ArgumentError, "The format of 'sprintf' is #{@operands[0].inspect}, which the fact doesn't have")
+    end
     formatted(
-      _values(0, fact, maps, fb)[0],
+      fmt,
       (1..(@operands.length - 1)).map { |i| _values(i, fact, maps, fb)&.first }
     )
   end
--- a/lib/factbase/terms/sprintf.rb
+++ b/lib/factbase/terms/sprintf.rb
@@ -26,10 +26,7 @@ class Factbase::Sprintf < Factbase::TermBase
     if fmt.nil?
       raise(ArgumentError, "The format of 'sprintf' is #{@operands[0].inspect}, which the fact doesn't have")
     end
-    formatted(
-      fmt,
-      (1..(@operands.length - 1)).map { |i| _values(i, fact, maps, fb)&.first }
-    )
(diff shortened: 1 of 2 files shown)


morphqdd:

yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2338](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #727, opened 2026-09-03
#713: let the empty string literal through the tokenizer

### LINKED ISSUE #713: The tokenizer refuses the empty string literal `""`
## What happens

https://github.com/yegor256/factbase/blob/master/lib/factbase/syntax.rb#L140

```ruby
raise(ArgumentError, 'String literal can\'t be empty') if t.length <= 2
```

`t` here is the quoted token, so `""` has length 2 and is refused outright:

```
(eq foo "") -> ArgumentError: String literal can't be empty
```

There is no way to say "this property is the empty string": `(eq foo "")`,
`(ends_with foo "")` and `(as x "")` all fail to parse. Facts can hold `''` perfectly well, since
`fact.rb` only checks the class, so the store can contain values the query language cannot name.
There is no workaround either, because `(concat)` is not a legal substitute for a literal in a
`starts_with` or `sprintf` format position.

## What should happen

Allow `""` and `''` as valid literals producing `''`. If a guard is wanted it belongs on the
symbol branch, not on the string branch.

morphqdd:
`""` was refused outright, because the length check counted the quotes, so
`(eq foo "")`, `(ends_with foo "")` and any format position needing an empty
string failed to parse. The check now only catches a token that is not a closed
literal at all.

A fact still cannot hold an empty value, `fact.rb` refuses that; this is about
naming the empty string in a query.

Closes #713

### DIFF
--- a/lib/factbase/syntax.rb
+++ b/lib/factbase/syntax.rb
@@ -137,7 +137,7 @@ class Factbase::Syntax
       if t.is_a?(Symbol)
         t
       elsif t.start_with?('\'', '"')
-        raise(ArgumentError, 'String literal can\'t be empty') if t.length <= 2
+        raise(ArgumentError, 'String literal is not closed') if t.length < 2
         t[1..-2]
       elsif t.match?(/^(\+|-)?[0-9]+$/)
         Integer(t, 10)
--- a/test/factbase/test_syntax.rb
+++ b/test/factbase/test_syntax.rb
@@ -22,6 +22,16 @@ class TestSyntax < Factbase::Test
     end
   end
 
+  def test_parses_empty_string_literal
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks for the fix — allowing the empty string literal through the tokenizer resolves #713, with parsing and matching tests.

@rultor merge

---

yegor256:
Nice fix — the length check was counting the quotes themselves, which made no sense. Merged, thanks.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4038](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #729, opened 2026-09-03
#710: concatenate all values of an operand

### LINKED ISSUE #710: `concat` keeps only the first value of each multi-valued operand
## What happens

README line 169 says `(concat v1 v2 v3 ...)` concatenates all `v`. Every property in a factbase
is a set of values, but `concat` keeps only the first one of each operand:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/concat.rb#L24

```ruby
(0..(@operands.length - 1)).map { |i| _values(i, fact, maps, fb)&.first }.join
```

A fact with `foo = ['a', 'b']`:

```
[["a"]]           <- (as z (concat foo)); "ab" expected
```

There is no way to concatenate a multi-valued property.

## What should happen

Join all values of each operand rather than only the first, or document the truncation.

morphqdd:
`concat` kept only the first value of each operand, so a multi-valued property
could not be concatenated at all, while the README says it concatenates all `v`.
All values of each operand are joined now.

Closes #710

### DIFF
--- a/lib/factbase/terms/concat.rb
+++ b/lib/factbase/terms/concat.rb
@@ -21,6 +21,6 @@ class Factbase::Concat < Factbase::TermBase
   # @param [Factbase] fb Factbase to use for sub-queries
   # @return [String] The concatenated string
   def evaluate(fact, maps, fb)
-    (0..(@operands.length - 1)).map { |i| _values(i, fact, maps, fb)&.first }.join
+    (0..(@operands.length - 1)).map { |i| _values(i, fact, maps, fb)&.join }.join
   end
 end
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks for the fix — joining all values of each operand instead of only the first is exactly what #710 and the README call for, with a test.

@rultor merge

---

yegor256:
Good catch — concatenating only the first value of each operand clearly didn't match what the README promises. Merged, thanks.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3998](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #730, opened 2026-09-03
#709: keep a fact whose property carries a new value

### LINKED ISSUE #709: `unique` rejects a fact whose property carries a value never seen before
## What happens

README line 217 says `(unique p1 p2 ...)` returns true if at least one property value has not
been seen yet. The code says the opposite:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/unique.rb#L28

```ruby
!@seen.intersect?(tuples).tap { tuples.each { |t| @seen << t } }
```

`intersect?` is true as soon as any tuple was seen before, so the fact is kept only when none of
its tuples is known. With three facts `foo=[1]`, `foo=[1,2]`, `foo=[3]`, the middle fact carries
a brand new value `2` and should pass by the documented rule, but it is rejected because of the
`1`:

```
[[1], [3]]        <- expected [[1], [1, 2], [3]] per README
```

## What should happen

Either match the README, keeping the fact when any tuple is new, or change the README to say
that every value must be new. As written the code and the spec disagree.

morphqdd:
The README says `(unique p1 p2 ...)` is true when at least one value has not been
seen yet, and false when the whole combination is a duplicate. The code returned
false as soon as any tuple was known, so a fact carrying a brand new value was
dropped because of an old one next to it.

Single-valued properties behave exactly as before.

Closes #709

### DIFF
--- a/lib/factbase/terms/unique.rb
+++ b/lib/factbase/terms/unique.rb
@@ -25,6 +25,6 @@ class Factbase::Unique < Factbase::TermBase
     vv = (0..(@operands.size - 1)).map { |i| _values(i, fact, maps, fb) }
     return false if vv.any?(nil)
     tuples = Enumerator.product(*vv).to_a
-    !@seen.intersect?(tuples).tap { tuples.each { |t| @seen << t } }
+    tuples.filter_map { |t| @seen.add?(t) }.any?
   end
 end
--- a/test/factbase/terms/test_unique.rb
+++ b/test/factbase/terms/test_unique.rb
@@ -20,6 +20,14 @@ class TestUnique < Factbase::Test
     assert(t.evaluate(fact('foo' => 1), [], Factbase.new))
   end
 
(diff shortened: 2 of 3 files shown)


yegor256:
Thanks for the fix — matching the README's "any tuple is new" rule for `unique`, and updating the one existing assertion that relied on the old, opposite behavior. Addresses #709.

@rultor merge

---

yegor256:
Thanks — matches the README's definition of `unique` for multi-valued properties much better now. Merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3978](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #731, opened 2026-09-03
#708: round Time the same way for every comparison

### LINKED ISSUE #708: `Compare` rounds `Time` to whole seconds for `eq` only, so `eq` and `lte` contradict each other
## What happens

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/compare.rb#L31

```ruby
l = l.floor if l.is_a?(Time) && @op == :==
...
  r = r.floor if r.is_a?(Time) && @op == :==
```

The truncation to whole seconds is gated on `@op == :==`. For `t = 2024-01-01T00:00:00.500Z`,
`eq` compares `00:00:00` with `00:00:00` and says true, while the four ordering operators see the
raw sub-second value:

```
(eq t 2024-01-01T00:00:00Z)  -> true
(lt t 2024-01-01T00:00:00Z)  -> false
(lte t 2024-01-01T00:00:00Z) -> false
(gt t 2024-01-01T00:00:00Z)  -> true
(gte t 2024-01-01T00:00:00Z) -> true
```

So the same pair is at once equal and strictly greater, and equal but not less than or equal.
`(or (lt a b) (eq a b))` and `(lte a b)` give different answers, and the ordering is not a total
order.

## What should happen

One rounding policy for all five comparison operators. Either floor `Time` operands in every
branch, so `eq`, `lte` and `gte` agree, or drop the floor from `eq` and compare exact times
everywhere.

morphqdd:
The truncation of `Time` to whole seconds was gated on `eq`, so the same pair
could be equal and strictly greater at once, and `(or (lt a b) (eq a b))`
disagreed with `(lte a b)`. All five operators floor `Time` now, so the ordering
is total again.

Closes #708

### DIFF
--- a/lib/factbase/terms/compare.rb
+++ b/lib/factbase/terms/compare.rb
@@ -28,9 +28,9 @@ class Factbase::Compare < Factbase::TermBase
     rights = _values(1, fact, maps, fb)
     return false if rights.nil?
     lefts.any? do |l|
-      l = l.floor if l.is_a?(Time) && @op == :==
+      l = l.floor if l.is_a?(Time)
       rights.any? do |r|
-        r = r.floor if r.is_a?(Time) && @op == :==
+        r = r.floor if r.is_a?(Time)
         _compare(l, r)
       end
     end
--- a/test/factbase/terms/test_compare.rb
+++ b/test/factbase/terms/test_compare.rb
@@ -17,6 +17,16 @@ class TestCompare < Factbase::Test
     refute(Factbase::Compare.new(:<, [4, 2]).evaluate(fact, [], Factbase.new), 'Expected 2 < 4 to be true')
   end
 
+  def test_rounds_time_the_same_way_for_every_operator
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks for the fix — flooring `Time` for every comparison operator, not just `eq`, restores a total order as requested in #708, with a test covering all five operators.

@rultor merge

---

yegor256:
Right catch — flooring `Time` only for `eq` broke the total ordering across operators. Merged, thanks.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3958](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #732, opened 2026-09-03
#707: keep the facts that have no value for the sorting property

### LINKED ISSUE #707: `Sorted#predict` drops every fact that has no value for the sorting property
## What happens

`Sorted#evaluate` returns `true` unconditionally, so `(sorted p expr)` is meant to be pure
post-processing: reorder the facts, keep everything `expr` matched. But `predict` is what
`Factbase::Query#each` actually uses to build the candidate set, and it filters:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/sorted.rb#L34

```ruby
.reject { |m| m[prop].nil? }
```

Any fact without a value for the sorting property is dropped before the sort. So `predict`
returns a strict subset of what `evaluate` accepts. The sibling post-processors `Inverted#predict`
and `Head#predict` do not filter.

Three facts, `a` with age 30, `b` with age 20, `c` with no age at all:

```
[["a"], ["b"], ["c"]]     <- (always)
[["b"], ["a"]]            <- (sorted age (always))   c is gone
[["c"], ["b"], ["a"]]     <- (inverted (always))     c is kept
[2]                       <- (agg (sorted age (always)) (count))
[3]                       <- (agg (always) (count))
```

The count inside an `agg` is wrong as a result.

## What should happen

`sorted` should keep the facts that have no value for the property, placing them at one end or
sorting with a nil-safe key, the same way `inverted` and `head` keep everything.

morphqdd:
`evaluate` returns true for every fact, so `sorted` is meant to reorder and keep
everything, but `predict` dropped the facts without a value for the property.
The candidate set was a strict subset of what the term accepts, and a `count`
inside an `agg` came out wrong. `inverted` and `head` keep everything.

Facts without the property now go to the end of the sorted list.

Closes #707

### DIFF
--- a/lib/factbase/terms/sorted.rb
+++ b/lib/factbase/terms/sorted.rb
@@ -30,9 +30,8 @@ class Factbase::Sorted < Factbase::TermBase
     raise(ArgumentError, "A symbol is expected as first argument of 'sorted'") unless prop.is_a?(Symbol)
     term = @operands[1]
     raise(ArgumentError, "A term is expected, but '#{term}' provided") unless term.is_a?(Factbase::Term)
-    fb.query(term, maps).each(fb, params).to_a
-      .reject { |m| m[prop].nil? }
-      .sort_by { |m| m[prop].first }
+    blank, valued = fb.query(term, maps).each(fb, params).to_a.partition { |m| m[prop].nil? }
+    (valued.sort_by { |m| m[prop].first } + blank)
       .map { |m| m.all_properties.to_h { |k| [k, m[k]] } }
   end
 end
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks for the fix — keeping facts with no value for the sorting property (appended after the sorted ones) matches what #707 asked for and matches `inverted`/`head` behavior, with a test.

@rultor merge

---

yegor256:
Good fix — `predict` was silently dropping candidates that `evaluate` would have accepted, which could skew an `agg` count. Merged, thanks.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (15) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3938](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #733, opened 2026-09-03
#711: leave a non-term operand alone while simplifying

### LINKED ISSUE #711: `and` and `or` raise an internal `NoMethodError` when an operand is not a term
## What happens

`And#simplify` and `Or#simplify` hand every operand to `Simplified#unique`, which calls
`simplify` on it unconditionally:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/simplified.rb#L23

Only `Factbase::Term` responds to `simplify`, so a symbol or a numeric literal operand blows up
inside the deduplication helper. `Factbase::Syntax#to_term` wraps that as `Broken`, so the user
sees an internal `NoMethodError` instead of a message naming the term and the bad operand:

```
(or foo (eq a 1)) -> Factbase::Syntax::Broken: undefined method 'simplify' for an instance of Symbol
(and (eq a 1) 5)  -> Factbase::Syntax::Broken: undefined method 'simplify' for an instance of Integer
(not foo)         -> parsed
```

The crash happens at parse time, before any fact is touched, and only for `and` and `or`.
`(not foo)` parses fine and later fails with a proper message.

## What should happen

`Simplified#unique` should leave a non-term operand alone, and `and` and `or` should refuse a
non-boolean operand with a clean `ArgumentError` naming the operator and the operand, the way
`assert`, `traced`, `join` and `empty` already do.

morphqdd:
`Simplified#unique` called `simplify` on every operand, and only `Factbase::Term`
has that method, so `(or foo (eq a 1))` and `(and (eq a 1) 5)` died at parse time
with an internal `NoMethodError` wrapped as `Broken`.

Non-term operands are left alone now. Both queries parse, and evaluation reports
the bad operand by name the way it does for `not`.

Closes #711

### DIFF
--- a/lib/factbase/terms/simplified.rb
+++ b/lib/factbase/terms/simplified.rb
@@ -20,7 +20,7 @@ class Factbase::Simplified
     strs = []
     ops = []
     @operands.each do |o|
-      o = o.simplify
+      o = o.simplify if o.is_a?(Factbase::Term)
       s = o.to_s
       next if strs.include?(s)
       strs << s
--- a/test/factbase/terms/test_and.rb
+++ b/test/factbase/terms/test_and.rb
@@ -5,6 +5,7 @@ require_relative '../../../lib/factbase/terms/and'
 # SPDX-FileCopyrightText: Copyright (c) 2024-2026 Yegor Bugayenko
 # SPDX-License-Identifier: MIT
 
+require_relative '../../../lib/factbase/syntax'
 require_relative '../../test__helper'
 
 # Test for the 'and' term.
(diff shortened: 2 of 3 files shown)


yegor256:
Thanks for the fix — leaving a non-`Term` operand alone in `Simplified#unique` resolves the internal `NoMethodError` from #711, with parsing and evaluation tests for both `and` and `or`.

@rultor merge

---

yegor256:
Thanks for fixing this and adding coverage for both the parse-time crash and the runtime error message for a bad operand. Merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3918](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #734, opened 2026-09-03
#721: count facts in the size attribute

### LINKED ISSUE #721: `ToXML#xml` marshals the whole factbase to fill a `size` attribute that is a byte count
## What happens

https://github.com/yegor256/factbase/blob/master/lib/factbase/to_xml.rb#L32

```ruby
meta = { version: Factbase::VERSION, size: @fb.export.size }
```

`@fb.export` is `Marshal.dump(@maps)`, so `.size` is the byte length of a binary blob, not a fact
count. Everywhere else in the API `size` means the number of facts. A three-fact factbase renders
three `f` elements and `@size="37"`:

```
== X: ToXML meta size attribute ==
fb.size=3  facts in xml=3  @size=37
```

Since fb0a0380 the body is built from `@fb.each.to_a`, so the whole `Marshal.dump`, an extra full
copy of the factbase in memory, is now done purely to compute this one number. `ToJSON` and
`ToYAML` emit no such attribute. It also makes `ToXML` unusable on a factbase that has no
`export`, such as the `Factbase::Light` handed to a transaction block.

The only test touching it, `test_meta_data_presence`, asserts the attribute exists but never
checks its value.

## What should happen

The attribute should be `@fb.size`, matching its name and the rest of the API, with no
`Marshal.dump`. If the byte length is really wanted it deserves a different name.

morphqdd:
The `size` attribute held the byte length of `Marshal.dump` of the whole
factbase, while `size` means a fact count everywhere else in the API. The dump
was made purely for that one number, and it made `ToXML` unusable on a factbase
that has no `export`, such as the one handed to a transaction block.

Closes #721

### DIFF
--- a/lib/factbase/to_xml.rb
+++ b/lib/factbase/to_xml.rb
@@ -29,7 +29,7 @@ class Factbase::ToXML
   # Convert the entire factbase into XML.
   # @return [String] The factbase in XML format
   def xml
-    meta = { version: Factbase::VERSION, size: @fb.export.size }
+    meta = { version: Factbase::VERSION, size: @fb.size }
     Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
       xml.fb(meta) do
         Factbase::Flatten.new(@fb.each.to_a, @sorter).it.each do |m|
--- a/test/factbase/test_to_xml.rb
+++ b/test/factbase/test_to_xml.rb
@@ -33,6 +33,13 @@ class TestToXML < Factbase::Test
     xml = Nokogiri::XML.parse(Factbase::ToXML.new(fb).xml)
     refute_empty(xml.xpath('/fb[@version]'))
     refute_empty(xml.xpath('/fb[@size]'))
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks for the fix — using `@fb.size` instead of `Marshal.dump`-ing the whole factbase is exactly what #721 asked for, and it's covered by a new test.

@rultor merge

---

yegor256:
Makes sense — `size` should mean fact count everywhere, and this also unblocks `ToXML` on factbases without `export`. Merged, thanks.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (9) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3898](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #735, opened 2026-09-03
#720: print a Time in ISO 8601 in JSON too

### LINKED ISSUE #720: `ToJSON` prints a `Time` with `Time#to_s`, unlike `ToXML` and `ToYAML`
## What happens

The three serializers render the same stored `Time` three different ways. `ToJSON` has no `Time`
handling at all, values go straight to `to_json`:

https://github.com/yegor256/factbase/blob/master/lib/factbase/to_json.rb#L98

```
stored (Marshal round trip): 2026-09-01T10:20:30.123456Z
to_xml : <when t="T">2026-09-01T10:20:30Z</when>
to_json: [{"when":"2026-09-01 10:20:30 UTC"}]
to_yaml: - when: 2026-09-01 10:20:30.123456000 Z
```

The factbase keeps full sub-second precision, `ToYAML` keeps it, `ToXML` truncates to whole
seconds but at least emits ISO 8601 with a `t="T"` marker. `ToJSON` emits Ruby's `Time#to_s`
form, which is not ISO 8601, is indistinguishable from a plain string, and loses the sub-second
part too. Anything consuming the JSON has to special-case a Ruby-specific format.

## What should happen

All three serializers should render a `Time` the same way, in ISO 8601, keeping the precision the
factbase actually stores, for instance `val.utc.iso8601(6)`.

morphqdd:
`ToJSON` had no `Time` handling, so a stored time came out as Ruby's
`Time#to_s`, which is not ISO 8601 and drops the sub-second part the factbase
keeps. It now prints the same ISO 8601 string `ToXML` prints, and both keep
microseconds, so the two agree with what is actually stored.

Closes #720

### DIFF
--- a/lib/factbase/to_json.rb
+++ b/lib/factbase/to_json.rb
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: MIT
 
 require 'json'
+require 'time'
 require_relative '../factbase'
 require_relative '../factbase/flatten'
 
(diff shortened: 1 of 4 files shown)


yegor256:
Thanks for the fix — aligning `ToJSON` and `ToXML` on `iso8601(6)` matches the resolution suggested in #720, with tests confirming both formats now agree.

@rultor merge

---

yegor256:
Good call keeping JSON and XML time rendering in agreement, microseconds included. Merged, thanks.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3878](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #736, opened 2026-09-03
#716: restore the check when a transaction is aborted

### LINKED ISSUE #716: `Factbase::Rules#txn` never restores `@check`, so a rolled-back transaction disables every rule
## What happens

https://github.com/yegor256/factbase/blob/master/lib/factbase/rules.rb#L48-L63

Line 51 sets `@check = later`, and line 55 restores `@check = before` — but line 55 sits inside
the block passed to `@fb.txn`, after `yield`. When the block raises `Factbase::Rollback`, the
documented way to abort, or any other exception, `yield` unwinds and line 55 never runs.

`@check` then stays a `Factbase::Rules::Later`, whose `it` returns immediately when `@uid` is
nil, so every later write through the same `Rules` instance is unchecked, silently and forever.

```
== R: Rules after a rolled-back transaction ==
before txn: The fact doesn't match the "(exists foo)" rule: [ bar: [1] ]
after  txn: NO EXCEPTION - rule is gone

== R2: same, but the block raises a plain error ==
txn raised: boom
after  txn: NO EXCEPTION - rule is gone
```

## What should happen

`@check` should be restored in an `ensure`, so an aborted transaction leaves rule enforcement
exactly as it was.

morphqdd:
`@check` was restored after `yield`, inside the transaction block, so a
`Factbase::Rollback` or any other exception skipped the line and left the
instance with a `Later` check that never enforces anything. Every later write
through the same `Rules` object went unchecked.

Restored in an `ensure` now.

Closes #716

### DIFF
--- a/lib/factbase/rules.rb
+++ b/lib/factbase/rules.rb
@@ -51,8 +51,11 @@ class Factbase::Rules
     @check = later
     @fb.txn do |fbt|
       churn = Factbase::Churn.new
-      yield(Factbase::Tallied.new(Factbase::Rules.new(fbt, @rules, @check, uid: @uid), churn))
-      @check = before
+      begin
+        yield(Factbase::Tallied.new(Factbase::Rules.new(fbt, @rules, @check, uid: @uid), churn))
+      ensure
+        @check = before
+      end
       unless churn.zero?
         fbt.query('(always)').each do |f|
           next unless later.include?(f)
--- a/test/factbase/test_rules.rb
+++ b/test/factbase/test_rules.rb
@@ -32,6 +32,30 @@ class TestRules < Factbase::Test
     end
   end
 
+  def test_keeps_checking_after_a_rollback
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks for the fix — restoring `@check` in `ensure` is the right shape for #716, and both the rollback and raised-error paths are tested.

@rultor merge

---

yegor256:
Moving the restore into `ensure` is exactly right — leaving `@check` unrestored after a rollback could have quietly disabled checks for a long time. Merged, thanks.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3858](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #737, opened 2026-09-03
#714: find the original copy by identity

### LINKED ISSUE #714: `LazyTaped#source_of` matches copies by value, so a transaction duplicates one of two equal facts
## What happens

`_track` stores `@copies[original] = copy` in a `compare_by_identity` hash, and `source_of` does
the reverse lookup:

https://github.com/yegor256/factbase/blob/master/lib/factbase/lazy_taped.rb#L24-L27

```ruby
@copies.key(copy)
```

`compare_by_identity` only changes how keys are compared. `Hash#key` searches the values with
`==`. Two facts holding the same properties produce two copies that are `==` to each other, so
`source_of(copy_of_b)` returns `original_a`.

At commit (`lib/factbase.rb:203-211`) the `added` loop does
`original = taped.source_of(b); garbage[original] = true`. Both modified copies resolve to the
same original, so only one original is put in `garbage`. Both copies are appended, the second
original is never removed, and the factbase grows by one fact holding a stale, unmodified copy.

```
== A: two identical facts modified in one txn ==
before: [{"foo" => [1]}, {"foo" => [1]}]
churn : 0i/0d/2a
after : size=3 [{"foo" => [1]}, {"foo" => [1], "bar" => [2]}, {"foo" => [1], "bar" => [2]}]

== C: two DIFFERENT facts modified in one txn (control) ==
churn : 0i/0d/2a
after : size=2 [{"foo" => [1], "bar" => [9]}, {"foo" => [2], "bar" => [9]}]
```

Root cause in plain Ruby, no factbase involved:

```
h.size = 2
h.key(cb).equal?(b) = false  (expected true)
h.key(cb).equal?(a) = true
```

## What should happen

Changing a property on N facts in a transaction should leave exactly N facts, each carrying the
change. `source_of` should find the original by object identity, for instance by keeping a second
identity-keyed copy-to-original map in `_track`, never by `==`.

### LINKED ISSUE #715: Deleting two equal facts inside a transaction removes only one of them
## What happens

https://github.com/yegor256/factbase/blob/master/lib/factbase/lazy_taped.rb#L90-L97

```ruby
@deleted.append(source_of(m).object_id) if r
```

`source_of` matches copies by value, so for two equal facts it returns the same original both
times. `@deleted` gets that one `object_id` twice, and `deleted` at `lazy_taped.rb:43-45` calls
`.uniq`, so the commit loop in `lib/factbase.rb:196-201` sees a single deletion.

```
--- D: delete two identical facts inside a txn ---
before: size=2 maps=[{"foo" => [1]}, {"foo" => [1]}]
churn=0i/1d/0a
after : size=1 maps=[{"foo" => [1]}]

--- control: same delete OUTSIDE a txn ---
after : size=0 maps=[]
```

With three equal facts only one is removed as well: `churn=0i/1d/0a`, `after: size=2`.

## What should happen

`query(...).delete!` inside a transaction should delete the same facts as outside one, and the
churn should report the real count.

morphqdd:
`@copies` is keyed by identity, but `Hash#key` searches the values with `==`, so
`source_of` returned the wrong original whenever two facts held equal
properties. Modifying both in one transaction put a single original in the
garbage set and the factbase grew by one stale fact; deleting both appended the
same object id twice and `uniq` collapsed it into one deletion.

A second identity-keyed map from copy to original is kept now, so the lookup is
by identity in both directions. This is the same root cause as #715, so both
cases are covered here and tested.

Closes #714
Closes #715

### DIFF
--- a/lib/factbase/lazy_taped.rb
+++ b/lib/factbase/lazy_taped.rb
@@ -14,6 +14,7 @@ class Factbase::LazyTaped
     @staged = []
     @copied = false
     @copies = {}.compare_by_identity
+    @sources = {}.compare_by_identity
     @inserted = []
     @deleted = []
     @added = []
@@ -23,7 +24,7 @@ class Factbase::LazyTaped
   # Returns nil if the base hasn't been copied yet or if the fact is new.
   def source_of(copy)
     return unless @copied
-    @copies.key(copy)
+    @sources[copy]
   end
 
   def copied?
@@ -141,6 +142,7 @@ class Factbase::LazyTaped
 
   def _track(copy, original)
     @copies[original] = copy
+    @sources[copy] = original
   end
 
   def _tape(map)
--- a/test/factbase/test_lazy_taped.rb
+++ b/test/factbase/test_lazy_taped.rb
@@ -233,6 +233,30 @@ class TestLazyTaped < Factbase::Test
     assert_equal(2, fb.query('(always)').each.to_a.first.foo)
   end
 
+  def test_modifies_two_equal_facts_in_txn
+    fb = Factbase.new
+    fb.insert.foo = 1
+    fb.insert.foo = 1
+    fb.txn do |fbt|
+      fbt.query('(exists foo)').each { |f| f.bar = 2 }
+    end
+    assert_equal(2, fb.size)
+    assert_equal(2, fb.query('(exists bar)').each.to_a.size)
+  end
+
+  def test_deletes_two_equal_facts_in_txn
+    fb = Factbase.new
+    fb.insert.foo = 1
+    fb.insert.foo = 1
+    assert_equal(
+      2,
+      fb.txn do |fbt|
+        fbt.query('(exists foo)').delete!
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks for the fix — an identity-keyed `@sources` map resolves the copy-to-original ambiguity from #714 directly, with tests for both the modify and delete cases.

@rultor merge

---

yegor256:
Nice fix, and appreciated that it covers both #714 and #715 with one identity-keyed map plus tests for both directions. Merged, thanks.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3834](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #738, opened 2026-09-03
#717: give a transaction its own cache

### LINKED ISSUE #717: `CachedFactbase#txn` shares its cache, so facts from a rolled-back transaction keep being returned
## What happens

https://github.com/yegor256/factbase/blob/master/lib/factbase/cached/cached_factbase.rb#L59-L63

The inner factbase built for the transaction gets the same `@cache` object as the outer one.
`insert` inside the transaction sets `@cache[:__dirty__] = true`, the first query inside the
transaction consumes that flag in `invalidate_if_dirty!` (`cached_query.rb:137`) and then stores
its in-transaction result under the query key.

When the transaction rolls back nothing clears the cache, and the dirty flag has already been
consumed, so the next query outside the transaction serves the cached array holding a fact that
was never committed:

```
== C: CachedFactbase keeps facts from a rolled-back transaction ==
origin size = 1
query says  = ["kept", "ghost"]
```

The factbase reports one fact and the query yields two.

## What should happen

A transaction should not write into the outer cache, either by getting its own cache or by
having the outer one cleared when the transaction ends, whether it committed or rolled back.

morphqdd:
The factbase built for a transaction shared the outer cache, so a query run
inside the transaction consumed the dirty flag and stored its result under the
query key. After a rollback the next query outside served that array, holding a
fact that was never committed.

The transaction gets its own cache now, and the outer one is dropped when the
transaction ends, committed or not.

Closes #717

### DIFF
--- a/lib/factbase/cached/cached_factbase.rb
+++ b/lib/factbase/cached/cached_factbase.rb
@@ -58,7 +58,9 @@ class Factbase::CachedFactbase
   # @return [Factbase::Churn] How many facts have been changed (zero if rolled back)
   def txn
     @origin.txn do |fbt|
-      yield(Factbase::CachedFactbase.new(fbt, @cache))
+      yield(Factbase::CachedFactbase.new(fbt, {}))
     end
+  ensure
+    @cache.clear
   end
 end
--- a/test/factbase/cached/test_cached_factbase.rb
+++ b/test/factbase/cached/test_cached_factbase.rb
@@ -84,4 +84,16 @@ class TestCachedFactbase < Factbase::Test
     end
     assert_equal(2, fb.query('(always)').each.to_a.size)
   end
+
+  def test_query_after_rollback_forgets_the_facts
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks for the fix — giving the transaction its own cache and clearing the outer one in `ensure` closes the rollback leak cleanly, with a test for it.

@rultor merge

---

yegor256:
Sharing the outer cache with a transaction was a subtle one to spot — giving the transaction its own cache and dropping it in `ensure` is the right fix. Merged, thanks.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3814](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #739, opened 2026-09-03
#718: keep a fact fresh until the index has seen it

### LINKED ISSUE #718: `IndexedFactbase#query` clears `@fresh` too early, so a later property write leaves the index stale
## What happens

https://github.com/yegor256/factbase/blob/master/lib/factbase/indexed/indexed_factbase.rb#L53-L59

`insert` puts the new fact's wrapper id into `@fresh`. `IndexedFact` skips `@idx.clear` on a
property write while the wrapper is in `@fresh` (`indexed_fact.rb:30`), which is only sound while
the index has not yet seen the fact.

But `query` clears `@fresh` at query-construction time, before `each` runs and before `predict`
feeds the index. So a fact inserted after a query object is created, but before that query is
iterated, is fed into the index without the property and is still marked fresh, so its later
property write does not invalidate the index. Every later query answers from the stale index.

`truth` below is the same data queried through a plain, unindexed `Factbase`:

```
== I: IndexedFactbase stale index ==
run 1 : 1
truth : 2
run 2 : 1
```

## What should happen

A fact should stop being fresh once the index has actually been fed with it. Clear `@fresh` where
the index is built and iterated, not where the query object is constructed, so a later property
write invalidates the index.

morphqdd:
`@fresh` was cleared when the query object was built, before the index was fed.
A fact inserted after that, but before the query ran, went into the index without
its properties and stayed marked fresh, so its later property write did not
invalidate the index and every later query answered from a stale one.

The set is cleared where the index is actually built and iterated now.

Closes #718

### DIFF
--- a/lib/factbase/indexed/indexed_factbase.rb
+++ b/lib/factbase/indexed/indexed_factbase.rb
@@ -52,10 +52,7 @@ class Factbase::IndexedFactbase
   # @param [Array<Hash>] maps Possible maps to use
   def query(term, maps = nil)
     term = to_term(term) if term.is_a?(String)
-    q = @origin.query(term, maps)
-    q = Factbase::IndexedQuery.new(q, @idx, self, @fresh)
-    @fresh.clear
-    q
+    Factbase::IndexedQuery.new(@origin.query(term, maps), @idx, self, @fresh)
   end
 
   # Run an ACID transaction.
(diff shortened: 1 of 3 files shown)


yegor256:
Thanks for the fix — clearing `@fresh` where the index is actually fed rather than at query construction is exactly what #718 called for, and it's covered by a new regression test.

@rultor merge

---

yegor256:
Good catch on the timing of when `@fresh` gets cleared relative to the index actually being built. Merged, thanks.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3794](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #740, opened 2026-09-03
#719: print a value XML cannot hold in Base64

### LINKED ISSUE #719: `ToXML#xml` copies control characters verbatim, producing XML that cannot be read back
## What happens

`Factbase::Fact` accepts any `String` value, including one holding a C0 control character.
`ToXML#to_str` just calls `val.to_s`:

https://github.com/yegor256/factbase/blob/master/lib/factbase/to_xml.rb#L56-L62

Nokogiri's builder escapes `<`, `&` and quotes, but it does not reject or escape control
characters, which XML 1.0 forbids in character data. The document that comes out is not
well-formed. With the value `"a" + 1.chr + "b"`:

```
== N: control character in a value ==
strict reparse FAILED: 4:18: FATAL: PCDATA invalid Char value 1
non-strict reparse text: "ab"
```

A strict reparse fails outright and a lenient one silently drops the byte, so the value changes.

The existing test `test_complex_rendering` covers only `"￿ < > & ' \""`, which Nokogiri
accepts, so this gap is untested.

## What should happen

`ToXML` should either refuse a value holding characters XML 1.0 cannot represent, or encode it,
for instance as base64 behind a distinct `t=` marker, so the output is always well-formed and the
value survives the round trip.

morphqdd:
A fact can hold a string with a C0 control character, and the builder copied it
into the document verbatim. XML 1.0 forbids those, so a strict reparse failed
and a lenient one dropped the byte, changing the value.

Such a value is printed in Base64 now and marked with `t="B"`, so the document is
always well-formed and the value survives.

Closes #719

### DIFF
--- a/lib/factbase/to_xml.rb
+++ b/lib/factbase/to_xml.rb
@@ -16,10 +16,16 @@ require_relative '../factbase/flatten'
 #  fb = Factbase.new
 #  puts Factbase::ToXML.new(fb).xml
 #
+# A value holding a character that XML 1.0 cannot represent is printed
+# in Base64 and marked with t="B", since printing it verbatim would make
+# the document not well-formed.
+#
 # Author:: Yegor Bugayenko (yegor256@gmail.com)
 # Copyright:: Copyright (c) 2024-2026 Yegor Bugayenko
 # License:: MIT
 class Factbase::ToXML
+  BAD = /[^\u0009\u000A\u000D\u0020-\uD7FF\uE000-\uFFFD\u{10000}-\u{10FFFF}]/
+
   # Constructor.
   def initialize(fb, sorter = '_id')
     @fb = fb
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks for the fix — small, focused, well tested (Base64 fallback for control characters, with a round-trip test).

@rultor merge

---

yegor256:
Thanks for tracking this down — printing the offending value as Base64 with `t="B"` is a clean way to keep the document well-formed without silently losing data. Merged.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3774](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #752, opened 2026-09-05
#741: keep a non-term operand inside the term

### LINKED ISSUE #741: A single non-term operand of `and` or `or` escapes as the query term and crashes with `NoMethodError`
## What happens

A single non-term operand of `and` or `or` escapes as the query term itself and then
receives `predict`, which literals do not implement:

```ruby
fb = Factbase.new
fb.query('(and 42)').each.to_a
```

```
NoMethodError: undefined method 'predict' for an instance of Integer
  lib/factbase/query.rb:49:in 'Factbase::Query#each'
```

`(or 42)` fails the same way. The error is raw, with no mention of the term or the
query, unlike every other operand-type failure in the gem.

The cause is in `simplify`:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/and.rb#L32

```ruby
unique = Factbase::Simplified.new(@operands).unique
return unique[0] if unique.size == 1
```

With one operand, `unique[0]` is the bare `42`, and `query.rb:49` calls
`@term.predict(...)` on it.

Two operands take a different path and report the problem properly:

```
(and (always) 42)  ->  "Boolean is expected, while Integer received from 42"
(not 42)           ->  "Boolean is expected, while Integer received from 42"
```

## What should happen

`(and 42)` and `(or 42)` should report the same `Boolean is expected, while Integer
received from 42` as their two-operand forms, instead of a raw `NoMethodError` from
the prediction stage.

morphqdd:
`simplify` returned the operand itself when only one was left, so `(and 42)` became the bare `42` and the query then called `predict` on an Integer. It only collapses now when the remaining operand is a term, so `(and 42)` reports `Boolean is expected, while Integer received from 42`, the same as `(and (always) 42)`.

Closes #741

### DIFF
--- a/lib/factbase/terms/and.rb
+++ b/lib/factbase/terms/and.rb
@@ -29,7 +29,7 @@ class Factbase::And < Factbase::TermBase
 
   def simplify
     unique = Factbase::Simplified.new(@operands).unique
-    return unique[0] if unique.size == 1
+    return unique[0] if unique.size == 1 && (unique[0].is_a?(Factbase::Term) || unique[0].is_a?(Factbase::TermBase))
     Factbase::Term.new(@op, unique)
   end
 end
--- a/lib/factbase/terms/or.rb
+++ b/lib/factbase/terms/or.rb
@@ -29,7 +29,7 @@ class Factbase::Or < Factbase::TermBase
 
   def simplify
     unique = Factbase::Simplified.new(@operands).unique
-    return unique[0] if unique.size == 1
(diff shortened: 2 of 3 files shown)


yegor256:
Thanks, merged. Good catch on the simplify collapsing to a non-term.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4930](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #753, opened 2026-09-05
#742: check that when gets terms

### LINKED ISSUE #742: `when` accepts a non-term operand and fails with a misleading "not defined" message
## What happens

`when` calls `evaluate` on its operands without checking they are terms:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/when.rb#L23

```ruby
def evaluate(fact, maps, fb)
  assert_args(2)
  return true unless @operands[0].evaluate(fact, maps, fb)
  @operands[1].evaluate(fact, maps, fb)
end
```

A literal in either position produces an internal error wrapped in a message that
blames the wrong thing:

```ruby
fb.query('(when 42 (always))').each.to_a
```

```
RuntimeError: Probably the term 'when' is not defined at (when 42 (always)):
undefined method 'evaluate' for an instance of Integer
```

`(when (always) 42)` fails the same way. The term is defined, so the message sends
the reader looking for a missing term instead of a bad operand.

This is the same defect that #711 fixed for `and` and `or`. `agg`, `assert`, `empty`
and `join` all check their term operands and raise a clear `ArgumentError`.

## What should happen

`when` should check both operands the way `agg` and `assert` do, and raise
`A term is expected, but '42' provided`.

morphqdd:
`when` called `evaluate` on its operands without checking them, so a literal produced `Probably the term 'when' is not defined`, which blames a missing term instead of the bad operand. It now raises `A term is expected, but '42' provided`, the way `agg` and `assert` do.

Closes #742

### DIFF
--- a/lib/factbase/terms/when.rb
+++ b/lib/factbase/terms/when.rb
@@ -22,6 +22,11 @@ class Factbase::When < Factbase::TermBase
   # @return [Boolean] True if first operand is false OR both are true
   def evaluate(fact, maps, fb)
     assert_args(2)
+    @operands.each do |o|
+      unless o.is_a?(Factbase::Term) || o.is_a?(Factbase::TermBase)
+        raise(ArgumentError, "A term is expected, but '#{o}' provided")
+      end
+    end
     return true unless @operands[0].evaluate(fact, maps, fb)
     @operands[1].evaluate(fact, maps, fb)
   end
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks, merged. Good, clearer error for a bad operand.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4910](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #754, opened 2026-09-05
#743: forget what unique saw on the previous run

### LINKED ISSUE #743: `unique` keeps its `@seen` set between runs, so the second `each` on the same query returns nothing
## What happens

The same query object returns different facts on a second run:

```ruby
fb = Factbase.new
fb.insert.then { |f| f.name = 'bob' }
q = fb.query('(unique name)')
q.each.to_a.size  # => 2
q.each.to_a.size  # => 0
```

`@seen` lives in the term and is never reset between iterations:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/unique.rb#L24

```ruby
@seen = Set.new if @seen.nil?
```

After the first pass every tuple is in `@seen`, so the second pass rejects everything.
`prev` keeps state in the same way, but there the value carried over is at least
observable in the result; here the query silently becomes empty.

This is not #709. That one is about which facts pass within a single run. This one is
about the second run of the same object returning nothing at all.

## What should happen

A query object should be repeatable. Running `each` twice on it should give the same
facts both times, so `unique` should start each iteration with an empty `@seen`.

morphqdd:
`unique` kept `@seen` in the term, so a second `each` on the same query object rejected everything and returned nothing. Terms get a `reset` that walks the operand tree, `unique` drops `@seen` in it, and `Query#each` calls it before iterating, so a query object gives the same facts every time.

Closes #743

### DIFF
--- a/lib/factbase/query.rb
+++ b/lib/factbase/query.rb
@@ -45,6 +45,7 @@ class Factbase::Query
   def each(fb = @fb, params = {})
     return to_enum(__method__, fb, params) unless block_given?
     yielded = 0
+    @term.reset if @term.respond_to?(:reset)
     params = params.transform_keys(&:to_s) if params.is_a?(Hash)
     maybe = @term.predict(@maps, fb, Factbase::Tee.new({}, params))
     maybe ||= @maps unless maybe.equal?(@maps)
--- a/lib/factbase/term.rb
+++ b/lib/factbase/term.rb
@@ -173,11 +173,10 @@ class Factbase::Term < Factbase::TermBase
   end
 
   # Forget what previous iterations of this term have seen.
-  # @return [nil]
+  # @return [Array, nil] The operands that were walked
   def reset
     @terms[@op].reset if @terms.key?(@op) && @terms[@op].respond_to?(:reset)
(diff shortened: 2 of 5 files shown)


yegor256:
Holding off on merging this one automatically: the "License Compliance" check reports state success but its own description says "Analysis timed out. Check FOSSA for updates." so it's not clear the scan actually ran. Everything else (tests, lint) is green and the diff itself looks fine. Requesting a human look before merge.

---

opusstudiohq-max:
Offered rather than competing: `prev` has exactly the shape this PR fixes for `unique`, and #876 reports it — `@prev.tap { @prev = _values(...) }` never clears, so the second run of one query hands the first fact the last value of the run before (`[nil, [5], [7]]` then `[[9], [5], [7]]`).

Since this PR introduces the `reset` protocol and the `@term.reset` call in `Query#each`, `Prev` needs only:

```ruby
def reset
  @prev = nil
  super
end
```

plus a two-run test like `test_repeats_on_the_second_run`. I stopped short of opening a separate PR for it because it would depend on this one's infrastructure and conflict with it; if you would rather I send it as a follow-up once this lands, say so and I will.

---
_Generated by [Claude Code](https://claude.com/claude-code)_

PR #756, opened 2026-09-05
#745: use an aggregate that exists in the examples

### LINKED ISSUE #745: Both `agg` examples in README use `(avg ...)`, which is not a term
## What happens

The README uses `(avg ...)` in both of its `agg` examples. There is no `avg` term.

The first one is the opening code sample of the section:

https://github.com/yegor256/factbase/blob/master/README.md#L140

```ruby
fb.query(
  '(eq salary (agg (eq dept $dept) (avg salary)))'
).each(Factbase.new, dept: ['eng']).to_a
```

Running it:

```
RuntimeError: Probably the term 'avg' is not defined at (avg salary):
undefined method 'avg' for an instance of Factbase::Term
```

The same expression appears again in the bullet list of `agg` examples further down.

There is no `lib/factbase/terms/avg.rb`. The aggregates that do exist are `count`,
`sum`, `max`, `min`, `first` and `nth`. The other two `agg` examples in the same list,
`(max age)` and `(max id)`, both work.

## What should happen

Either add the `avg` term, or change both README examples to an aggregate that exists.
As written, the first executable sample a reader meets for `agg` does not run.

morphqdd:
Both `agg` examples in the README used `(avg salary)`, and there is no `avg` term, so the first executable sample a reader meets for `agg` raised `Probably the term 'avg' is not defined`. Both now use `(max salary)`, and a test runs that example.

Closes #745


yegor256:
Holding off on merging this one automatically: the "License Compliance" check reports state success but its own description says "Analysis timed out. Check FOSSA for updates." so it's not clear the scan actually ran. Everything else (tests, lint) is green and the diff itself looks fine. Requesting a human look before merge.

---

PR #757, opened 2026-09-05
#746: say which value cannot be summed

### LINKED ISSUE #746: `sum` over a non-numeric property leaks a raw coercion error naming neither property nor value
## What happens

`sum` adds every value with `+` and lets Ruby's coercion error out:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/sum.rb#L34

```ruby
vv.each do |v|
  sum += v
end
```

Over a string property:

```ruby
fb = Factbase.new
fb.query('(gt (sum name) 0)').each.to_a
```

```
RuntimeError: "String can't be coerced into Integer" at (sum name)
at lib/factbase/terms/sum.rb:34:in 'Integer#+'
```

A `Time` property gives `Time can't be coerced into Integer` the same way.

The message names neither the property nor the offending value, so on a factbase with
many facts there is no way to tell which one is not a number. Every other term that
needs a specific type says what it wanted: `matches` raises `Exactly one string is
expected`, `nth` raises `An integer is expected, but ... provided`, `to_integer` goes
through `Integer()` with the value in the message.

## What should happen

`sum` should either skip non-numeric values or raise a message naming the property and
the value that is not a number, instead of leaking `Integer#+`'s coercion error.

morphqdd:
`sum` added every value with `+` and let Ruby's `String can't be coerced into Integer` out, naming neither the property nor the value, so on a large factbase there was no way to tell which fact was wrong. It now raises `The value 'alice' of the 'name' property is not a number, can't sum it up`.

Closes #746

### DIFF
--- a/lib/factbase/terms/sum.rb
+++ b/lib/factbase/terms/sum.rb
@@ -31,6 +31,9 @@ class Factbase::Sum < Factbase::TermBase
       next if vv.nil?
       vv = [vv] unless vv.respond_to?(:to_a)
       vv.each do |v|
+        unless v.is_a?(Integer) || v.is_a?(Float)
+          raise(ArgumentError, "The value '#{v}' of the '#{k}' property is not a number, can't sum it up")
+        end
         sum += v
       end
     end
--- a/test/factbase/terms/test_sum.rb
+++ b/test/factbase/terms/test_sum.rb
@@ -41,7 +41,10 @@ class TestSum < Factbase::Test
   def test_names_the_property_and_the_value
     fb = Factbase.new
     fb.insert.name = 'alice'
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks, merged. Good improvement for debugging bad data.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4866](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #758, opened 2026-09-05
#747: report which two values cannot be compared

### LINKED ISSUE #747: `max` and `min` over a property with mixed types leak a raw comparison error
## What happens

`max` and `min` compare values with `>` and `<` and let Ruby error out:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/max.rb#L13

```ruby
fb = Factbase.new
fb.insert.then { |f| f.v = 's' }
fb.insert.then { |f| f.v = 1 }
fb.query('(gt (max v) 0)').each.to_a
```

```
RuntimeError: "comparison of Integer with String failed" at (max v)
at lib/factbase/terms/max.rb:13:in 'Integer#>'
```

Nothing in the message says which property or which fact holds the odd type, and a
factbase can legitimately carry a property whose type differs between facts, since
`Fact` validates each value on its own and never checks it against earlier facts.

## What should happen

`max` and `min` should report the property and the two values that could not be
compared, the way `nth` and `matches` report their type failures.

morphqdd:
`max` and `min` compared with `>` and `<` and let `comparison of Integer with String failed` out, which says nothing about the property or the values. The comparison is wrapped now and reports `Can't compare '1' (Integer) with 's' (String) in the 'v' property`.

Closes #747

### DIFF
--- a/lib/factbase/terms/best.rb
+++ b/lib/factbase/terms/best.rb
@@ -17,7 +17,18 @@ class Factbase::Best
       next if vv.nil?
       vv = [vv] unless vv.respond_to?(:to_a)
       vv.each do |v|
-        best = v if best.nil? || @criteria.call(v, best)
+        if best.nil?
+          best = v
+          next
+        end
+        begin
+          best = v if @criteria.call(v, best)
+        rescue ArgumentError, TypeError
+          raise(
+            ArgumentError,
+            "Can't compare '#{v}' (#{v.class}) with '#{best}' (#{best.class}) in the '#{key}' property"
+          )
+        end
       end
     end
     best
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks, merged. Nice, informative error message.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4846](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #759, opened 2026-09-05
#748: let a fact hold a boolean

### LINKED ISSUE #748: `_values` allows `TrueClass` and `FalseClass`, which `Fact` refuses to store
## What happens

The type whitelist that operands are validated against includes booleans:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/base.rb#L62

```ruby
unless v.all? { |i| [Float, Integer, String, Time, TrueClass, FalseClass].any? { |t| i.is_a?(t) } }
  raise 'Wrong type inside'
end
```

`Fact` cannot store either of them:

```ruby
raise(ArgumentError, "The type '#{v.class}' of '#{kk}' is not allowed") unless [
  String, Integer, Float,
  Time
].include?(v.class)
```

```ruby
fb.insert.then { |f| f.b = true }
```

```
ArgumentError: The type 'TrueClass' of 'b' is not allowed
```

So no property array can ever hold a `TrueClass` or `FalseClass`, and the two entries
in the `_values` whitelist can never be reached from a property. The two lists are the
only places where the set of allowed scalar types is written down, and they disagree.

## What should happen

The two lists should say the same thing. Either `Fact` should accept booleans, or
`_values` should drop them from its whitelist.

morphqdd:
The operand whitelist in `_values` allowed `TrueClass` and `FalseClass` while `Fact` refused both, so the two lists that spell out the allowed scalar types disagreed. `Fact` takes booleans now, which is the side that can change: dropping them from `_values` breaks `and` and `or`, whose operands are terms that evaluate to booleans.

Closes #748

### DIFF
--- a/lib/factbase/fact.rb
+++ b/lib/factbase/fact.rb
@@ -57,7 +57,7 @@ class Factbase::Fact
       raise(ArgumentError, "The value of '#{kk}' can't be empty") if v == ''
       raise(ArgumentError, "The type '#{v.class}' of '#{kk}' is not allowed") unless [
         String, Integer, Float,
-        Time
+        Time, TrueClass, FalseClass
       ].include?(v.class)
       v = v.utc if v.is_a?(Time)
       @map[kk] = [] if @map[kk].nil?
--- a/test/factbase/test_fact.rb
+++ b/test/factbase/test_fact.rb
@@ -105,4 +105,16 @@ class TestFact < Factbase::Test
     f.foo = 42
     assert_includes(f.all_properties, 'foo')
   end
+
+  def test_stores_booleans
+    fb = Factbase.new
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks, merged. Makes sense to align Fact with the operand whitelist.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4826](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #760, opened 2026-09-05
#749: say the input is not a factbase

### LINKED ISSUE #749: `import` guards the empty input but leaks raw `Marshal` errors for corrupt or truncated data
## What happens

`import` checks for empty input and reports it well, then hands everything else to
`Marshal` unguarded:

```ruby
fb = Factbase.new
fb.import('')
```

```
StandardError: Empty input, cannot load a factbase
```

```ruby
fb.import('not a factbase')
```

```
TypeError: incompatible marshal file format (can't be read)
        format version 4.8 required; 110.111 given
```

```ruby
good = Factbase.new
good.insert.then { |f| f.a = 1 }
data = good.export
fb = Factbase.new
fb.import(data[0..(data.size / 2)])
```

```
ArgumentError: marshal data too short
```

Neither of the last two mentions a factbase, so a caller that reads bytes from a file
gets an error naming an implementation detail it never chose. `judges` is such a
caller: `Judges::Impex#import` does `fb.import(File.binread(@file))` and passes the
error straight to the user, who is told `marshal data too short` about a file whose
path is not in the message.

Truncated input is not a hypothetical. `Judges::Impex#export` writes the factbase in
place, so a process killed mid-write leaves exactly the half-written file that produces
the third error above.

The empty-input guard shows the intended shape of the message. The other two corrupt
cases just do not have one.

## What should happen

`import` should catch `TypeError` and `ArgumentError` from `Marshal` and re-raise them
as a factbase-level error saying the input is not a valid factbase, the way the empty
case already does.

morphqdd:
`import` guarded empty input but let `Marshal`'s own errors out for corrupt or truncated bytes, so a caller reading a half-written file got `marshal data too short` with nothing about a factbase in it. Both are wrapped now, the way the empty case already was.

Closes #749

### DIFF
--- a/lib/factbase.rb
+++ b/lib/factbase.rb
@@ -253,12 +253,12 @@ class Factbase
   # @param [String] bytes Binary string to import
   def import(bytes)
     raise(StandardError, 'Empty input, cannot load a factbase') if bytes.empty?
-    data = Marshal.load(bytes)
-    @maps +=
-      if data.is_a?(Hash) && data.key?(:maps)
-        Marshal.load(data[:maps])
-      else
-        data
-      end
+    begin
+      data = Marshal.load(bytes)
+      data = Marshal.load(data[:maps]) if data.is_a?(Hash) && data.key?(:maps)
+    rescue TypeError, ArgumentError => e
+      raise(StandardError, "The input is not a valid factbase: #{e.message}")
+    end
+    @maps += data
   end
 end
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks, merged. Good defensive fix for corrupt input.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4806](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #761, opened 2026-09-05
#750: say a duration needs a whole number

### LINKED ISSUE #750: A fractional duration such as `1.5 hours` fails with a raw `Kernel#Integer` error, while a bad unit is reported properly
## What happens

The duration parser validates the unit but not the amount:

```ruby
fb = Factbase.new
fb.insert.then { |f| f.when = Time.now - 3600 }
fb.query("(gt when (minus (to_time '2026-09-04T07:40:32Z') '1 fortnight'))").each.to_a
```

```
RuntimeError: "Unknown time unit 'fortnight' in '1 fortnight'" at (minus ...)
```

That is a clear message. The amount gets none:

```ruby
fb.query("(gt when (minus (to_time '2026-09-04T07:40:32Z') '1.5 hours'))").each.to_a
```

```
RuntimeError: "invalid value for Integer(): \"1.5\"" at (minus ...) at <internal:kernel>:290:in 'Kernel#Integer'
```

```ruby
fb.query("(gt when (minus (to_time '2026-09-04T07:40:32Z') 'abc hours'))").each.to_a
```

```
RuntimeError: "invalid value for Integer(): \"abc\"" at (minus ...)
```

Neither says that a duration needs a whole number, or shows the duration string that
was rejected. `1.5 hours` is a reasonable thing to write and reads as valid, so the
error a user gets is about `Kernel#Integer` rather than about durations.

Callers build these strings by interpolation, so the value often comes from data rather
than from the query author. `Fbe.repeatedly` in `zerocracy/fbe` builds
`'#{hours} hours'` straight from a factbase fact, so a fact holding `1.5` turns into
this error at runtime.

## What should happen

The amount should be validated the way the unit already is, raising something like
`Amount '1.5' in '1.5 hours' must be a whole number`, so the message names the duration
rather than the conversion that failed.

morphqdd:
A duration like `1.5 hours` failed with `invalid value for Integer(): "1.5"`, which says nothing about durations and does not show the string that was rejected. It now raises `Amount '1.5' in '1.5 hours' must be a whole number`, next to the unit check that was already there.

Closes #750

### DIFF
--- a/lib/factbase/terms/arithmetic.rb
+++ b/lib/factbase/terms/arithmetic.rb
@@ -33,7 +33,11 @@ class Factbase::Arithmetic < Factbase::TermBase
     r = rights[0]
     if v.is_a?(Time) && r.is_a?(String)
       (num, units) = r.split
-      num = Integer(num, 10)
+      begin
+        num = Integer(num, 10)
+      rescue ArgumentError, TypeError
+        raise(ArgumentError, "Amount '#{num}' in '#{r}' must be a whole number")
+      end
       r =
         case units
         when 'seconds', 'second'
--- a/test/factbase/terms/test_minus.rb
+++ b/test/factbase/terms/test_minus.rb
@@ -35,4 +35,15 @@ class TestMinus < Factbase::Test
     )
     assert_nil(t.evaluate(fact, [], Factbase.new))
   end
+
+  def test_rejects_a_fractional_duration
+    fb = Factbase.new
+    fb.insert.when = Time.now
+    ['1.5 hours', 'abc hours'].each do |d|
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks, merged. Much clearer error message.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4786](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #762, opened 2026-09-05
#751: print the term, not the object graph

### LINKED ISSUE #751: `Query#one` and `Query#each` build error messages with `@term.inspect`, producing megabyte-long messages
## What happens

`Query#one` and `Query#each` build their error messages with `@term.inspect`:

https://github.com/yegor256/factbase/blob/master/lib/factbase/query.rb#L74

```ruby
raise(StandardError, "Incorrect type #{r.class} returned by #{@term.inspect}")
```

https://github.com/yegor256/factbase/blob/master/lib/factbase/query.rb#L57

```ruby
raise(ArgumentError, "Unexpected evaluation result of type #{r.class}, must be Boolean at #{@term.inspect}")
```

`Factbase::Term` carries `@terms`, the table of every registered term class, so
`inspect` walks all of them. On the simplest possible term the message is already 5 KB:

```ruby
fb = Factbase.new
fb.insert.then { |f| f.what = 'issue' }
fb.query('(eq what "issue")').one
```

```
message bytes=4959
first 120: Incorrect type FalseClass returned by #<Factbase::Term:0x00007546d340a850 @op=:eq, @operands=[:what, "issue"], @terms={u
```

On a three-clause `(and ...)` used through `Fbe.iterate`, the same raise produced a
message of 1.1 MB, which fills a CI log with one exception.

`Term` already renders itself readably, and every other error in the gem uses that
form:

```ruby
q.instance_variable_get(:@term).to_s   # => "(eq what 'issue')"
```

```
RuntimeError: "..." at (eq (sprintf $absent 'hi') 'hi') at ...
```

So the two `inspect` calls are the only places that dump the internals instead of the
expression.

## What should happen

Both messages should interpolate the term directly, the way the rest of the gem does,
so the reader sees `(eq what 'issue')` rather than several kilobytes of object graph.

morphqdd:
The two error messages in `query.rb` used `@term.inspect`, which dumps the whole object graph: a five-kilobyte message for `(eq what 'issue')` and 1.1 MB for a three-clause `and`. They interpolate the term itself now, like the rest of the gem does.

Closes #751

### DIFF
--- a/lib/factbase/query.rb
+++ b/lib/factbase/query.rb
@@ -54,7 +54,7 @@ class Factbase::Query
       f = Factbase::Tee.new(f, params)
       r = @term.evaluate(Factbase::Accum.new(f, extras, false), @maps, fb)
       unless r.is_a?(TrueClass) || r.is_a?(FalseClass)
-        raise(ArgumentError, "Unexpected evaluation result of type #{r.class}, must be Boolean at #{@term.inspect}")
+        raise(ArgumentError, "Unexpected evaluation result of type #{r.class}, must be Boolean at #{@term}")
       end
       next unless r
       yield(Factbase::Accum.new(f, extras, true))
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks, merged. Good catch on the object-graph dump.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4766](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #773, opened 2026-09-06
#767: accept a float the way Ruby writes it

### LINKED ISSUE #767: `1.5e10` is rejected, because the float pattern demands a sign in the exponent
## What happens

The tokenizer accepts a float only if it has a dot and, when it has an exponent, a sign
in it:

```ruby
elsif t.match?(/^(\+|-)?[0-9]+\.[0-9]+(e(\+|-)[0-9]+)?$/)
  Float(t)
```

So three ordinary float literals are refused:

```ruby
fb.query('(eq n 1.5e10)').each.to_a
```

```
Factbase::Syntax::Broken: Wrong symbol format (1.5e10)
```

The same for `1.5E10` and for `1e10`. Only `1.5e+10` goes through.

`Float()` in Ruby, and every other language with this syntax, accepts all four. The error
message points at a symbol, which sends the reader looking for a typo in a property name
rather than at the number.

## What should happen

The pattern should accept a float the way Ruby writes it, with an optional sign in the
exponent, an upper case `E`, and no dot when an exponent is present.

morphqdd:
The float branch of the tokenizer demanded a dot and, when an exponent was present, a
sign in it, so `1.5e10`, `1.5E10` and `1e10` were all refused as a wrong symbol format
while `1.5e+10` went through.

The pattern now takes either a decimal with an optional exponent, or an integer with an
exponent, with the sign optional and either case of `e`. The integer branch above it
still catches a plain integer first, so nothing that used to parse as an integer becomes
a float.

Closes #767


yegor256:
This now conflicts with master (a couple of sibling PRs touching the same regex in syntax.rb landed first), so it needs a rebase before it can be merged. Leaving this for a manual look.

---

PR #774, opened 2026-09-06
#768: give a parameter name the same rule as a property name

### LINKED ISSUE #768: A query parameter cannot carry a digit in its name, while a property can
## What happens

A property name may hold digits and capitals, a parameter name may not:

```ruby
raise(ArgumentError, "Wrong symbol format (#{t})") unless t.match?(/^([_a-z][a-zA-Z0-9_]*|\$[_a-z]+)$/)
```

The property branch is `[_a-z][a-zA-Z0-9_]*`, the parameter branch is `\$[_a-z]+`.

```ruby
fb.query('(eq id2 $id2)').each(fb, 'id2' => [7]).to_a
```

```
Factbase::Syntax::Broken: Wrong symbol format ($id2)
```

`(eq id2 $id)` works, so the same name is legal on the left and illegal on the right.
The natural thing to write, when a parameter carries the value of a property, is to give
it the property's name, and that is exactly what breaks.

## What should happen

A parameter name should follow the same rule as a property name, `\$[_a-z][a-zA-Z0-9_]*`.

morphqdd:
The symbol pattern let a property name hold digits and capitals (`[_a-z][a-zA-Z0-9_]*`)
but a parameter name only lowercase letters and underscores (`\$[_a-z]+`), so `(eq id2 $id)`
parsed and `(eq id2 $id2)` did not.

Both branches are now one pattern with an optional leading `$`.

Closes #768

### DIFF
--- a/lib/factbase/syntax.rb
+++ b/lib/factbase/syntax.rb
@@ -146,7 +146,7 @@ class Factbase::Syntax
       elsif t.match?(/^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$/)
         Time.parse(t)
       else
-        raise(ArgumentError, "Wrong symbol format (#{t})") unless t.match?(/^([_a-z][a-zA-Z0-9_]*|\$[_a-z]+)$/)
+        raise(ArgumentError, "Wrong symbol format (#{t})") unless t.match?(/^\$?[_a-z][a-zA-Z0-9_]*$/)
         t.to_sym
       end
     end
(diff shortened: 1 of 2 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5860](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #775, opened 2026-09-06
#769: take a time literal with fractional seconds

### LINKED ISSUE #769: A time literal with fractional seconds is rejected, though the gem prints times that way
## What happens

The tokenizer takes a time literal only without fractional seconds:

```ruby
elsif t.match?(/^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$/)
  Time.parse(t)
```

```ruby
fb.query("(gt when 2026-09-05T07:00:00.123456Z)").each.to_a
```

```
Factbase::Syntax::Broken: Wrong symbol format (2026-09-05T07:00:00.123456Z)
```

Without the fraction the same literal works.

The gem prints its own times with six fractional digits:

https://github.com/yegor256/factbase/blob/master/lib/factbase/to_xml.rb

```ruby
val.utc.iso8601(6)
```

and `ToJSON` does the same. So a time taken from the gem's own output cannot be pasted
back into a query, and a caller that interpolates `Time#utc.iso8601(6)` into a query gets
a broken one.

## What should happen

The pattern should accept the fractional part, since it is the format the gem itself
produces.

morphqdd:
The time pattern in the tokenizer had no room for the fractional part, so
`2026-09-05T07:00:00.123456Z` was refused as a wrong symbol format. That is the shape
`ToXML` and `ToJSON` write (`iso8601(6)`), so a time taken out of the gem's own output
could not be pasted back into a query.

The fraction is now optional in the pattern; `Time.parse` already reads it.

Closes #769

### DIFF
--- a/lib/factbase/syntax.rb
+++ b/lib/factbase/syntax.rb
@@ -143,7 +143,7 @@ class Factbase::Syntax
         Integer(t, 10)
       elsif t.match?(/^(\+|-)?[0-9]+\.[0-9]+(e(\+|-)[0-9]+)?$/)
         Float(t)
-      elsif t.match?(/^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$/)
+      elsif t.match?(/^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(\.[0-9]+)?Z$/)
         Time.parse(t)
       else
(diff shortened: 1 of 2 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5840](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #776, opened 2026-09-06
#770: list a shared property name once

### LINKED ISSUE #770: `Factbase::Tee#all_properties` lists a name twice when the fact and the params share it
## What happens

`Factbase::Tee#all_properties` concatenates two lists without removing what they share:

https://github.com/yegor256/factbase/blob/master/lib/factbase/tee.rb

```ruby
def all_properties
  (@fact.is_a?(Hash) ? @fact.keys : @fact.all_properties) +
    (@upper.is_a?(Hash) ? @upper.keys : @upper.all_properties)
end
```

```ruby
fb = Factbase.new
f = fb.insert
f.foo = 1
f.bar = 2
Factbase::Tee.new(f, { 'foo' => [9] }).all_properties
# => ["foo", "bar", "foo"]
```

`Factbase::Accum`, which sits next to it and does the same job for a different pair of
sources, uses `|`:

https://github.com/yegor256/factbase/blob/master/lib/factbase/accum.rb

```ruby
def all_properties
  @fact.all_properties | @props.keys
end
```

Every fact yielded by `Query#each` is wrapped into a `Tee` with the query parameters, so
a query whose parameter is named after a property gives every consumer of
`all_properties` the same name twice.

## What should happen

`Tee#all_properties` should return each name once, the way `Accum#all_properties` does.

morphqdd:
`Tee#all_properties` concatenated the fact's names with the params' names, so a name held
by both was listed twice. `Accum#all_properties`, which does the same job for a different
pair of sources, uses `|`; this does the same.

Closes #770

### DIFF
--- a/lib/factbase/tee.rb
+++ b/lib/factbase/tee.rb
@@ -27,7 +27,7 @@ class Factbase::Tee
   end
 
   def all_properties
-    (@fact.is_a?(Hash) ? @fact.keys : @fact.all_properties) +
+    (@fact.is_a?(Hash) ? @fact.keys : @fact.all_properties) |
       (@upper.is_a?(Hash) ? @upper.keys : @upper.all_properties)
   end
 
--- a/test/factbase/test_tee.rb
+++ b/test/factbase/test_tee.rb
@@ -68,8 +68,7 @@ class TestTee < Factbase::Test
     prim = Factbase::Fact.new({})
     prim.foo = 42
     prim.bar = 13
-    t = Factbase::Tee.new(prim, { 'foo' => [9] })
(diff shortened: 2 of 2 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (9) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5820](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #777, opened 2026-09-06
#771: rebuild a fact only from the properties it has

### LINKED ISSUE #771: `sorted`, `head` and `inverted` turn a query parameter into a property with no value
## What happens

`sorted`, `head` and `inverted` rebuild every fact as a plain hash, from the list of its
property names:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/sorted.rb

```ruby
(valued.sort_by { |m| m[prop].first } + blank)
  .map { |m| m.all_properties.to_h { |k| [k, m[k]] } }
```

The facts they get come from `Query#each`, which wraps each one into a `Factbase::Tee`
together with the query parameters, and `Tee#all_properties` includes the parameter
names. So a parameter turns into a property with no value:

```ruby
fb = Factbase.new
%w[bob alice].each { |n| fb.insert.name = n }
fb.query('(sorted name (eq name $who))').each(fb, who: %w[alice bob]).to_a.each do |f|
  puts "#{f.all_properties.inspect} who=#{f['who'].inspect}"
end
```

```
["name", "n", "who"] who=nil
["name", "n", "who"] who=nil
```

`all_properties` lists `who`, and reading `who` gives nil. Nothing else in the gem
produces a fact like that: `Factbase::Fact` refuses to set a property to nil, so a
property that exists always has at least one value. Any consumer that walks
`all_properties` and reads each name, `Fbe.overwrite` and `Fbe.copy` among them, breaks
on the nil.

`head` and `inverted` have the same line.

## What should happen

These three terms should carry the facts through as they are, or rebuild them only from
the properties the fact itself has, so that a parameter never becomes a property.

morphqdd:
`sorted`, `head` and `inverted` rebuilt every fact from `all_properties`. The facts they
get come out of `Query#each` wrapped in a `Factbase::Tee` together with the query
parameters, so a parameter name ended up in the map with a nil value, which is not
something a property may hold.

The rebuilding is now one place, `_flatten` in `TermBase`, and it skips a name that reads
back as nil. It also removes the same two lines from three terms.

Closes #771

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -62,6 +62,7 @@ Elegant/GoodMethodName:
     - '_by_symbol'
     - '_compare'
     - '_feed'
+    - '_flatten'
     - '_init_indexes'
     - '_join'
     - '_resolve'
--- a/lib/factbase/terms/base.rb
+++ b/lib/factbase/terms/base.rb
@@ -37,6 +37,23 @@ class Factbase::TermBase
     raise(ArgumentError, "Too few (#{c}) operands for '#{@op}' (#{num} expected)") if c < num
   end
 
+  # Turns facts into plain maps, keeping only the properties they really carry.
+  #
+  # A fact coming out of a query is a +Factbase::Tee+, whose +all_properties+
+  # also lists the names of the query parameters. Reading such a name back gives
+  # NIL, which is not a value any property may have.
+  #
+  # @param [Array<Factbase::Fact>] facts The facts to turn into maps
+  # @return [Array<Hash>] The maps
+  def _flatten(facts)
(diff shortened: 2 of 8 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5800](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #779, opened 2026-09-06
#764: print a string of bad bytes in Base64

### LINKED ISSUE #764: `Factbase::ToXML` raises on a string that is not valid UTF-8, though it promises Base64 for it
## What happens

The class doc of `Factbase::ToXML` says:

https://github.com/yegor256/factbase/blob/master/lib/factbase/to_xml.rb#L19

```
A value holding a character that XML 1.0 cannot represent is printed
in Base64 and marked with t="B", since printing it verbatim would make
the document not well-formed.
```

That works for a character that is valid UTF-8 but illegal in XML, a NUL for example.
It does not work for a string whose bytes are not valid UTF-8:

```ruby
fb = Factbase.new
fb.insert.title = "bad\xFF".dup.force_encoding('UTF-8')
Factbase::ToXML.new(fb).xml
```

```
ArgumentError: invalid byte sequence in UTF-8
```

The `BAD` regular expression is matched against the value before anything else, and the
match itself raises, so the Base64 branch below is never reached.

`Factbase::Fact` checks the class of a value, never its encoding:

so such a string is storable, and it arrives in a factbase whenever a byte sequence from
an external source is put into a fact.

## What should happen

A string that is not valid UTF-8 should take the same Base64 path as a string with an
illegal character, since the intent of that path is exactly to hold what XML cannot.

morphqdd:
The `BAD` match was the first thing done to a value, and matching a regular expression
against bytes that are not valid UTF-8 raises `ArgumentError` itself, so a string that
the Base64 branch exists for never reached it.

The encoding is now checked before the match, and such a string takes the same `t="B"`
path as a string with a character XML cannot hold.

Closes #764

### DIFF
--- a/lib/factbase/to_xml.rb
+++ b/lib/factbase/to_xml.rb
@@ -64,7 +64,7 @@ class Factbase::ToXML
   # @param [Symbol] name The name of the element
   # @param [Object] val The value
   def put(xml, name, val)
-    if val.is_a?(String) && val.match?(BAD)
+    if val.is_a?(String) && (!val.valid_encoding? || val.match?(BAD))
       xml.__send__(name, [val].pack('m0'), t: 'B')
     else
       xml.__send__(name, to_str(val), t: type_of(val))
--- a/test/factbase/test_to_xml.rb
+++ b/test/factbase/test_to_xml.rb
@@ -32,6 +32,16 @@ class TestToXML < Factbase::Test
     assert_equal("a#{1.chr}b", node.text.unpack1('m0'))
   end
 
+  def test_broken_utf8_rendering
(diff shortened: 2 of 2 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5756](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #780, opened 2026-09-06
#765: name the property that JSON cannot hold

### LINKED ISSUE #765: `Factbase::ToJSON` raises on a string that `Factbase::Fact` accepts
## What happens

`Factbase::ToJSON` fails on a string that `Factbase::Fact` accepts:

```ruby
fb = Factbase.new
fb.insert.title = "bad\xFF".dup.force_encoding('UTF-8')
Factbase::ToJSON.new(fb).json
```

```
JSON::GeneratorError: source sequence is illegal/malformed utf-8
```

`Factbase::Fact` validates the class of a value and never its encoding:

`Factbase::ToXML` has an escape hatch for a value it cannot print verbatim, it prints it
in Base64 and marks it with `t="B"`. `ToJSON` has nothing of the sort, so the same
factbase can be printed as XML and cannot be printed as JSON.

`judges print --format=json` is the caller, and it reports the raw `JSON::GeneratorError`
without naming the fact or the property that carries the bad value.

## What should happen

`ToJSON` should either encode such a value the way `ToXML` does, or fail with a message
naming the property and the fact, instead of leaking the generator's own error.

morphqdd:
`Factbase::Fact` accepts a string whatever its encoding, so a value that is not valid
UTF-8 can sit in a factbase, and `ToJSON` then leaked `JSON::GeneratorError` with nothing
in it about where the bad value came from.

JSON has no place to mark an encoded value, the way XML has `t="B"`, so this takes the
other option the issue lists: the failure now names the property and the fact.

Closes #765

### DIFF
--- a/lib/factbase/to_json.rb
+++ b/lib/factbase/to_json.rb
@@ -30,16 +30,27 @@ class Factbase::ToJSON
   # @return [String] The factbase in JSON format
   def json
     Factbase::Flatten.new(@fb.each.to_a, @sorter).it.map do |m|
-      m.transform_values { |vv| vv.is_a?(Array) ? vv.map { |v| plain(v) } : plain(vv) }
+      m.to_h do |k, vv|
+        [k, vv.is_a?(Array) ? vv.map { |v| plain(m, k, v) } : plain(m, k, vv)]
+      end
     end.to_json
   end
 
   private
 
   # Render one value the way ToXML renders it.
+  # @param [Hash] map The fact the value belongs to
+  # @param [String] key The name of the property
   # @param [Object] val The value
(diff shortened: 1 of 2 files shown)


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5736](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #781, opened 2026-09-06
#766: refuse to define or undefine a built-in term

### LINKED ISSUE #766: `defn` and `undef` silently ignore every term implemented as a class
## What happens

`defn` refuses to touch a term that already exists:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/defn.rb#L20

```ruby
raise(ArgumentError, "Can't use '#{fn}' name as a term") if Factbase::Term.method_defined?(fn)
raise(ArgumentError, "Term '#{fn}' is already defined") if Factbase::Term.private_method_defined?(fn, false)
```

Both guards look for a METHOD on `Factbase::Term`. Every built-in term is a class listed
in `@terms`, not a method, so the guards never fire for any of them:

```ruby
fb = Factbase.new
fb.insert.foo = 42
fb.query('(defn eq "false")').each.to_a.size   # => 1, reported as success
fb.query('(eq foo 42)').each.to_a.size         # => 1, the old term is still there
```

The redefinition is silently ignored, because `evaluate` looks at `@terms` first:

https://github.com/yegor256/factbase/blob/master/lib/factbase/term.rb#L203

```ruby
if @terms.key?(@op)
  @terms[@op].evaluate(fact, maps, fb)
```

`undef` has the same hole:

```ruby
fb.query('(undef eq)').each.to_a.size   # => 1, reported as success
fb.query('(eq foo 42)').each.to_a.size  # => 1, still working
```

The README says `(undef f)` "undefines a term". All 60 built-ins (`eq`, `and`, `plus`,
`agg` and the rest) cannot be undefined or redefined, and neither call says so.

## What should happen

`defn` should refuse a name that `@terms` already holds, with the same "Term is already
defined" error it gives for a method, and `undef` should either remove the class-based
term or say that it cannot.

morphqdd:
Both guards in `defn` looked for a method on `Factbase::Term`, and every built-in term is
a class in the map that `evaluate` reads first, not a method. So `(defn eq ...)` and
`(undef eq)` both answered success and changed nothing.

The map of built-ins is now a constant, `Factbase::Term::TERMS`, built once from name to
class instead of name to instance, and `defn` and `undef` check it. `undef` says it
cannot remove such a term rather than pretending it did.

Closes #766


yegor256:

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +9 for the 131 hits-of-code that you wrote; -10 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5696](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #782, opened 2026-09-06
#763: make the term only when it is used

### LINKED ISSUE #763: `Factbase::Term` builds all sixty term objects for every node of every query
## What happens

The constructor of `Factbase::Term` builds every term class it knows about, for every node
of every query:

https://github.com/yegor256/factbase/blob/master/lib/factbase/term.rb#L102

```ruby
@terms = {
  unique: Factbase::Unique.new(operands),
  prev: Factbase::Prev.new(operands),
  ...
  max: Factbase::Max.new(operands)
}
```

There are 60 entries. Only one of them, `@terms[@op]`, is ever read, in `evaluate`,
`predict` and `simplify`. The other 59 are built and thrown away.

Measured on master:

```ruby
a = GC.stat(:total_allocated_objects)
Factbase::Term.new(:eq, [:foo, 42])
GC.stat(:total_allocated_objects) - a   # => 253
```

```ruby
a = GC.stat(:total_allocated_objects)
Factbase::Syntax.new('(and (eq foo 42) (gt bar 1))').to_term
GC.stat(:total_allocated_objects) - a   # => 406

20_000.times { Factbase::Syntax.new('(and (eq foo 42) (gt bar 1))').to_term }
# => 5.42s
```

A query is parsed on every `fb.query('...')` call, and judges call it per fact, so this
is on the hot path.

## What should happen

The term object should be built for the operator that is actually used, not for all of
them. A prototype that keeps a frozen `operator => class` map and builds the one term
lazily gives 9 allocations per `Term.new`, 150 per parse, and 2.06s for the same 20k
parses, with all 600 tests passing.

morphqdd:
The constructor built all sixty term classes for every node of every query, and only the
one under `@op` was ever read.

The instance is made on first use and kept. `Factbase::Term.new(:eq, [:foo, 42])` goes from
253 allocations to 7, and 20k parses of `(and (eq foo 42) (gt bar 1))` from 5.42s to 0.69s
here.

Stacked on #781, which turns the map of instances into a map of classes; the two lines that
read `@terms` outside this file, in `IndexedTerm#predict` and in the README test, follow it.

Closes #763


yegor256:
This is stacked on #781, which just got merged, so the branch here now conflicts with master and needs a rebase before it can go in. Leaving this for a manual look.

---

morphqdd:
Rebased on master, the conflict is gone and CI is green. @yegor256, please take another look.

PR #801, opened 2026-09-07
#787: delete the facts the query yields, not the ones it evaluates

### LINKED ISSUE #787: `delete!` on a `head` query deletes every fact, not the ones it matches
## What happens

`delete!` deletes facts that the same query does not match:

```ruby
fb = Factbase.new
5.times { |i| f = fb.insert; f.foo = i }
fb.query('(head 2 (exists foo))').each.to_a.map(&:foo)   # => [0, 1]
fb.query('(head 2 (exists foo))').delete!                # => 5
fb.size                                                  # => 0
```

`(head 0 (exists foo))` deletes all five too.

`each` narrows the list through `predict`:

https://github.com/yegor256/factbase/blob/master/lib/factbase/terms/head.rb

```ruby
def evaluate(_fact, _maps, _fb)
  true
end

def predict(maps, fb, params)
  ...
  .take(max)
```

and `delete!` only calls `evaluate`:

https://github.com/yegor256/factbase/blob/master/lib/factbase/query.rb

```ruby
def delete!(fb = @fb)
  deleted = 0
  @maps.delete_if do |m|
    d = @term.evaluate(Factbase::Accum.new(Factbase::Fact.new(m), {}, false), @maps, fb)
```

which is `true` for every fact.

I ran the same comparison for every term in the README: `head` is the only one where the
two disagree. `sorted` and `inverted` also answer `true` from `evaluate`, but they return
every fact from `each` as well, so nothing is lost there.

`one` has the same root: `fb.query('(head 1 (exists foo))').one` raises
`Incorrect type TrueClass returned by (head 1 (exists foo))`.

## What should happen

`delete!` should delete exactly the facts that `each` returns, so it should go through
`predict` the way `each` does.

morphqdd:
`delete!` had two paths. When every fact carries an id it collects the ids from `each`, which is
right. When one does not, it fell back to running the term against every fact, and `head`,
`sorted` and `inverted` all evaluate to `true` for everything: they narrow in `predict`, not in
`evaluate`. So `(head 2 ...)` deleted all five facts, and `(head 0 ...)` did too.

The fallback narrows through `predict` first now, the way `each` does, and takes each predicted
fact once, so two identical facts under `(head 1 ...)` lose one of them rather than both.

The predicted list is copied before it is walked: `predict` can hand back the very array that is
being deleted from, and consuming entries out of it while `delete_if` runs over it skipped facts.

Closes #787

### DIFF
--- a/lib/factbase/query.rb
+++ b/lib/factbase/query.rb
@@ -88,8 +88,11 @@ class Factbase::Query
   # @return [Integer] Total number of facts deleted
   def delete!(fb = @fb)
     deleted = 0
+    maybe = (@term.predict(@maps, fb, Factbase::Tee.new({}, {})) || @maps).to_a.dup
     @maps.delete_if do |m|
-      d = @term.evaluate(Factbase::Accum.new(Factbase::Fact.new(m), {}, false), @maps, fb)
+      pos = maybe.index(m)
+      d = !pos.nil? && @term.evaluate(Factbase::Accum.new(Factbase::Fact.new(m), {}, false), @maps, fb)
+      maybe.delete_at(pos) if d
       deleted += 1 if d
       d
     end
--- a/test/factbase/test_query.rb
+++ b/test/factbase/test_query.rb
@@ -161,6 +161,12 @@ class TestQuery < Factbase::Test
     assert_equal(1, maps.size)
   end
 
+  def test_deletes_only_the_head
+    maps = [{ 'foo' => [0] }, { 'foo' => [1] }, { 'foo' => [2] }, { 'foo' => [3] }, { 'foo' => [4] }]
+    assert_equal(2, Factbase::Query.new(maps, '(head 2 (exists foo))', Factbase.new).delete!)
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks, merged! Confirming each candidate against what the term predicts is the right shape, and the new test pins `head` down.

One thing worth a follow-up rather than a change here: `maybe.index(m)` is a linear scan inside `delete_if`, so a large `delete!` is now quadratic. A hash keyed by `object_id` would keep it linear.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6880](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #802, opened 2026-09-07
#788: refuse a float that is not a number

### LINKED ISSUE #788: `NaN` and `Infinity` are accepted by a fact, then break `ToJSON`, `sorted` and every indexed comparison
## What happens

`Factbase::Fact` accepts a non-finite float, since it only checks the class of a value:

```ruby
f = fb.insert
f.value = Float::NAN        # accepted
f.value = Float::INFINITY   # accepted
```

Nothing else in the gem can then deal with it.

`ToJSON` fails:

```
JSON::GeneratorError: NaN not allowed in JSON
JSON::GeneratorError: Infinity not allowed in JSON
```

so `judges print --format json` on such a factbase reports the generator's own error,
naming neither the fact nor the property.

Queries fail as well, as soon as an index is in front of the factbase:

```
plain   (gt num 0): 2 facts
plain   (sorted num (always)): ArgumentError: comparison of Float with NaN failed
indexed (gt num 0): ArgumentError: comparison of Float with NaN failed
indexed (lt num 5): ArgumentError: comparison of Float with NaN failed
```

One NaN in one fact makes every `gt`/`gte`/`lt`/`lte` query raise, because the index sorts
the values, and `sorted` raises even without an index. The plain factbase quietly leaves
the fact out of the answer instead.

`ToXML` writes `<value t="F">Infinity</value>`, which is not the `xs:double` lexical form
(`INF`), though Saxon happens to accept it.

This is the numeric twin of #765, which is about a string that is not valid UTF-8. The fix
there checks the encoding, so it will not cover this.

## What should happen

Either `Factbase::Fact` should refuse NaN and Infinity the way it refuses a Symbol, or
every consumer should handle them: `ToJSON` with a message naming the property, the index
without raising, and `ToXML` with `INF`.

morphqdd:
A fact only checked the class of a value, so `Float::NAN` and `Float::INFINITY` went in, and
nothing afterwards could deal with them. `ToJSON` raised `NaN not allowed in JSON`, naming neither
the fact nor the property, so `judges print --format json` failed with the generator's own
message. `sorted` and every comparison under an index raised `comparison of Float with NaN
failed`, again naming nothing.

The value is refused at the door instead, next to the checks for nil and for an empty string, so
the message names the property the way those do.

Closes #788

### DIFF
--- a/lib/factbase/fact.rb
+++ b/lib/factbase/fact.rb
@@ -59,6 +59,7 @@ class Factbase::Fact
         String, Integer, Float,
         Time, TrueClass, FalseClass
       ].include?(v.class)
+      raise(ArgumentError, "The value of '#{kk}' can't be #{v}") if v.is_a?(Float) && !v.finite?
       v = v.utc if v.is_a?(Time)
       @map[kk] = [] if @map[kk].nil?
       @map[kk] << v
--- a/test/factbase/test_fact.rb
+++ b/test/factbase/test_fact.rb
@@ -13,6 +13,13 @@ require_relative '../test__helper'
 # Copyright:: Copyright (c) 2024-2026 Yegor Bugayenko
 # License:: MIT
 class TestFact < Factbase::Test
+  def test_refuses_a_float_that_is_not_a_number
+    f = Factbase::Fact.new({})
+    [Float::NAN, Float::INFINITY, -Float::INFINITY].each do |v|
+      assert_raises(StandardError) { f.value = v }
+    end
+  end
+
   def test_injects_data_correctly
     map = {}
     f = Factbase::Fact.new(map)


yegor256:
Merged, thanks! Refusing it at the door, beside the nil and empty-string checks, is the right place — the error now names the property instead of coming out of the JSON generator.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+7220](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #803, opened 2026-09-07
#785: give the facts back in the order they went in

### LINKED ISSUE #785: `IndexedFactbase` returns `gt`, `gte`, `lt` and `lte` results in value order, not in insertion order
## What happens

The same factbase and the same query give the facts in a different order, depending on
whether an index is in front of it.

```ruby
def build(fb)
  [5, 1, 3, 2].each_with_index do |n, i|
    f = fb.insert
    f.num = n
    f.tag = "t#{i}"
  end
  fb
end

plain = build(Factbase.new)
indexed = build(Factbase::IndexedFactbase.new(Factbase.new))
```

```
(lt num 4)
  plain:   ["t1", "t2", "t3"]
  indexed: ["t1", "t3", "t2"]
(gt num 1)
  plain:   ["t0", "t2", "t3"]
  indexed: ["t3", "t2", "t0"]
```

The plain factbase yields in insertion order, the indexed one in the order of the compared
value. The set of facts is the same, only the order differs.

The reason is in the index itself:

https://github.com/yegor256/factbase/blob/master/lib/factbase/indexed/indexed_gt.rb

```ruby
def _feed(facts, entry, prop)
  ...
```

`_search` then returns the tail of that sorted array, so the caller gets the facts sorted
by `prop`. `IndexedGte`, `IndexedLt` and `IndexedLte` are written the same way.

I compared the plain, the indexed and the cached factbase on 10 random factbases of 30
facts against 36 queries. The three agree on which facts match every time; the only
differences are this ordering, for `gt`, `gte`, `lt` and `lte`, and for an `or` of two
`eq` terms.

It matters because `Fbe.fb` puts `IndexedFactbase` into the stack in production, so
`fb.query('(gt when ...)').each.first` is one fact in a test with a plain factbase and
another one in a real run. `sorted` and `head` exist for callers who want a specific
order, and they still work, but a caller who does not use them cannot tell what order it
will get.

## What should happen

An indexed query should return the facts in the same order a plain one does, so that the
index changes the speed of a query and nothing else.

morphqdd:
The index keeps its pairs sorted by the compared value so that it can binary-search them, and it
handed the matches back in that order. A plain factbase yields in insertion order, so the same
query gave a different order depending on whether an index sat in front of it, and a judge that
takes the first match got a different fact in production than in a test.

Each pair carries the position of its fact now, and the matched slice is put back into that order
before the facts are handed out. The pairs stay sorted by value, so the search is the same.

Closes #785

### DIFF
--- a/lib/factbase/indexed/indexed_gt.rb
+++ b/lib/factbase/indexed/indexed_gt.rb
@@ -32,9 +32,9 @@ class Factbase::IndexedGt
 
   def _feed(facts, entry, prop)
     return unless entry[:count] < facts.size
-    facts[entry[:count]..].each do |fact|
+    facts[entry[:count]..].each_with_index do |fact, i|
       fact[prop]&.each do |v|
-        entry[:facts] << [v, fact]
+        entry[:facts] << [v, fact, entry[:count] + i]
       end
     end
     entry[:facts].sort_by! { |pair| pair[0] }
@@ -42,9 +42,9 @@ class Factbase::IndexedGt
   end
 
   def _search(entry, target)
-    idx = entry[:facts].bsearch_index { |v, _| v > target }
+    idx = entry[:facts].bsearch_index { |pair| pair[0] > target }
     return [] if idx.nil?
-    facts = entry[:facts][idx..].map { |_, f| f }
+    facts = entry[:facts][idx..].sort_by { |pair| pair[2] }.map { |pair| pair[1] }
     facts.uniq!(&:object_id)
     facts
   end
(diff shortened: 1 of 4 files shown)


yegor256:
Thanks, merged! Carrying the insertion position in the index entry and sorting the slice by it is the clean way to get insertion order back, and it also makes the result independent of `sort_by!` not being stable.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6904](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #804, opened 2026-09-07
#786: name the term when an index cannot compare

### LINKED ISSUE #786: Comparison of mixed types under an index raises a bare `ArgumentError`, not the descriptive failure
## What happens

Comparing values of different types gives a descriptive failure without an index and a
raw one with it:

```ruby
def build(fb)
  fb.insert.num = 5
  fb.insert.num = 'alpha'
  fb
end

build(Factbase.new).query('(gt num 2)').each.to_a
build(Factbase::IndexedFactbase.new(Factbase.new)).query('(gt num 2)').each.to_a
```

```
plain:   RuntimeError: "Cannot compare \"alpha\" (String) with 2 (Integer) using (compare >):
         comparison of String with 2 failed" at (gt num 2)
indexed: ArgumentError: comparison of Integer with String failed
```

The indexed message says neither which property nor which query it came from, and the pair
of types it names is not the pair the query compares: it comes from sorting the index, so
it is one stored value against another, not a value against the target.

https://github.com/yegor256/factbase/blob/master/lib/factbase/indexed/indexed_gt.rb

```ruby
```

`sort_by!` raises `ArgumentError` straight from Ruby, and nothing above it turns that into
the message `Factbase::Term#evaluate` produces for the same data.

`IndexedGte`, `IndexedLt` and `IndexedLte` sort the same way.

## What should happen

A query that cannot compare its values should fail with the same message whether an index
is in front of the factbase or not.

morphqdd:
`Term#evaluate` wraps whatever goes wrong inside it with the term it happened in, which is why the
plain path says `"comparison of String with 2 failed" at (gt num 2)`. `predict` had no such
wrapper, and the indexed path raises there, when it sorts the pairs it wants to binary-search, so
the failure escaped as a bare `ArgumentError` naming neither the property nor the query.

`IndexedTerm#predict` wraps the same way `evaluate` does. The message still names the two stored
values rather than a value and the target, because that is what actually failed, but the term next
to it says which property and which query it came from.

Closes #786

### DIFF
--- a/lib/factbase/indexed/indexed_term.rb
+++ b/lib/factbase/indexed/indexed_term.rb
@@ -40,6 +40,10 @@ module Factbase::IndexedTerm
     return __send__(m, maps, fb, params) if respond_to?(m)
     _init_indexes unless @indexes
     @indexes[@op].predict(maps, fb, params) if @indexes.key?(@op)
+  rescue NoMethodError => e
+    raise(RuntimeError, "Probably the term '#{@op}' is not defined at #{self}: #{e.message}")
+  rescue StandardError => e
+    raise(RuntimeError, "#{e.message.inspect} at #{self} at #{e.backtrace[0]}")
   end
 
   private
--- a/test/factbase/indexed/test_indexed_term.rb
+++ b/test/factbase/indexed/test_indexed_term.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative '../../../lib/factbase'
+require_relative '../../../lib/factbase/indexed/indexed_factbase'
 require_relative '../../../lib/factbase/indexed/indexed_term'
 require_relative '../../../lib/factbase/taped'
 require_relative '../../../lib/factbase/term'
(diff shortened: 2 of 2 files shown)


yegor256:
CI is green and the added context is genuinely useful, but I'd rather not merge this one unreviewed, so I'm assigning it.

Two things I'd want a second opinion on:

The `rescue StandardError` arm rewrites *every* failure inside `predict` into a `RuntimeError`. #786 is about a mixed-type comparison surfacing a bare `ArgumentError`, which is one class; catching all of `StandardError` also flattens any Factbase-specific error that a caller may currently rescue by type, and it does so silently, since the original class survives only as text in the message.

`e.backtrace[0]` is read without a guard. A re-raised or manually constructed exception can carry a `nil` backtrace, and then the rescue itself raises `NoMethodError` on `nil` — inside the handler, so the original error is lost entirely.

Narrowing the second rescue to the comparison errors actually being reported, and guarding the backtrace read, would make this one safe to take.

---

PR #808, opened 2026-09-07
#800: Name the failed clause in a long conjunction rule

### LINKED ISSUE #800: A failed rule is reported by its first 32 characters, which for a rule read from a file is the license header
## What happens

`Factbase::Rules::Check#it` truncates the rule to its first 32 characters when it complains:

```ruby
raise(
  ArgumentError,
  "The fact doesn't match the #{(@expr.length > 32 ? "#{@expr[0..31]}..." : @expr).inspect} rule: #{fact}"
)
```

Any rule long enough to be worth naming is longer than 32 characters, and a rule assembled from a file starts with that file's header. So the message quotes the header and stops.

## Reproduction

```ruby
require 'factbase'
require 'factbase/rules'

long = "(and\n" + (1..6).map { |i| "# a comment line number #{i} that is long enough to matter\n" }.join + "  (exists foo)\n)"
fb = Factbase::Rules.new(Factbase.new, long)
fb.txn { |t| f = t.insert; f.bar = 1 }
```

```
ArgumentError: The fact doesn't match the "(and\n# a comment line number 1 t..." rule: [ bar: [1] ] + {}
```

The clause that actually failed is `(exists foo)`, three hundred characters further down, and nothing in the message points at it.

This is not a contrived shape. `zerocracy/fbe` builds its rule by reading `rules/*.fe` and wrapping the lot in one `(and ...)`, and those files open with a twenty-one line MIT header, so every violation in a judges run reports:

```
The fact doesn't match the "(and \n# MIT License\n#\n# Copyrigh..." rule: [ _id: [1], ... ]
```

Every rule failure in that project, whatever the cause, produces the same sentence about the MIT license.

## What would help

Two things, either of which is an improvement on its own.

Name the clause that failed rather than the head of the whole expression. `Check#it` already has the parsed term, so it can walk the operands of a top-level `(and ...)` and report the first one that evaluates false:

```ruby
def it(fact, fb)
  term = Factbase::Syntax.new(@expr).to_term
  return if term.evaluate(fact, [], fb)
  failed = term.op == :and ? term.operands.find { |o| !o.evaluate(fact, [], fb) } : term
  raise(ArgumentError, "The fact doesn't match #{failed}: #{fact}")
end
```

If the whole expression must still be shown, take the tail rather than the head, or strip comment lines before truncating, since the head of a file is the least informative part of it.

Note that `fbe/rules/basic.fe` defines its own `explain` term that logs which rule failed, which suggests the missing detail was felt there and worked around outside this class.

gemshrine:
Fixes #800.

`Factbase::Rules::Check` truncated every rule at 32 characters before putting
it in a violation. For a rule loaded from a file, those first characters are
the SPDX/license header, so the exception showed the header instead of the
failed rule. A long top-level conjunction also hid which clause was false.

The check now parses the expression once and, for a top-level `and`, reports
the first operand that evaluates to false. Short non-conjunction rules retain
their complete expression, while a long conjunction names the failing clause
rather than an arbitrary prefix.

The regression uses a long file-style header followed by `(exists foo)` and
asserts that the violation contains the failed clause but not the header.

Tests:

- `bundle exec rubocop lib/factbase/rules.rb test/factbase/test_rules.rb` — passed.

### DIFF
--- a/lib/factbase/rules.rb
+++ b/lib/factbase/rules.rb
@@ -128,11 +128,17 @@ class Factbase::Rules
     end
 
     def it(fact, fb)
-      return if Factbase::Syntax.new(@expr).to_term.evaluate(fact, [], fb)
-      raise(
-        ArgumentError,
-        "The fact doesn't match the #{(@expr.length > 32 ? "#{@expr[0..31]}..." : @expr).inspect} rule: #{fact}"
-      )
+      term = Factbase::Syntax.new(@expr).to_term
+      return if term.evaluate(fact, [], fb)
+      text = (@expr.length > 32 ? "#{@expr[0..31]}..." : @expr).inspect
+      failed = Check.failed(term, fact, fb)
+      text = "#{failed} of the #{text}" unless failed.nil? || failed.to_s == term.to_s
+      raise(ArgumentError, "The fact doesn't match the #{text} rule: #{fact}")
+    end
+
+    def self.failed(term, fact, fb)
+      return term unless term.op == :and
+      term.operands.find { |operand| !operand.evaluate(fact, [], fb) }
     end
   end
 
--- a/test/factbase/test_rules.rb
+++ b/test/factbase/test_rules.rb
@@ -169,4 +169,23 @@ class TestRules < Factbase::Test
       'Error message should truncate long expressions'
     )
   end
+
+  def test_error_message_names_failed_part_of_long_conjunction
+    f = Factbase::Rules.new(
+      Factbase.new,
+      <<~RULE
+        (and
+          # The rule is intentionally long enough that its beginning is not useful.
+          # This is how a rule read from a file normally begins.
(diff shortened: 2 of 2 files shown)


yegor256:
Reviewed and I am happy with the approach — naming the failing conjunct beats truncating the rule text. It cannot be merged as it stands, though: the branch now conflicts with `master` on `lib/factbase/rules.rb`. Please merge `master` in and I will take another look.

---

gemshrine:
Rebased on master, the conflict is gone. Master meanwhile pinned the truncation in `test_error_message_truncates_long_expression`, so the message now keeps both parts: it names the failing conjunct and still shows the truncated rule text, e.g. `The fact doesn't match the (exists foo) of the "(and\n  # The rule is intent..." rule: ...`. Both that test and the new one for the long conjunction pass.

PR #810, opened 2026-09-07
#798: Pass outer query parameters to agg selectors

### LINKED ISSUE #798: A query parameter inside agg is resolved against the current fact, so the README example returns the wrong facts
## What happens

The README shows a query parameter used inside `agg`:

```ruby
fb.query(
  '(eq salary (agg (eq dept $dept) (avg salary)))'
).each(Factbase.new, dept: ['eng']).to_a
```

> selects the facts with the salary average in their departments

The parameter never reaches the sub-query. `Factbase::Agg#evaluate` ends with

```ruby
term.evaluate(nil, fb.query(selector, maps).each(fb, fact).to_a, fb)
```

and the second argument of `Query#each` is the params hash. What is passed there is `fact`, the fact currently being evaluated. So inside the selector a `$name` is resolved against the current fact's properties, and whatever the caller passed to `each` is gone.

The result depends on whether the parameter happens to share its name with a property, which makes it hard to spot.

## Reproduction

Three facts: `dept=eng salary=100`, `dept=eng salary=200`, `dept=ops salary=50`. The correct answer to "the top salary in dept eng" is the single fact with salary 200.

```ruby
fb.query('(eq salary (agg (eq dept "eng") (max salary)))').each(fb).to_a
fb.query('(eq salary (agg (eq dept $dept) (max salary)))').each(fb, dept: ['eng']).to_a
fb.query('(eq salary (agg (eq dept $d) (max salary)))').each(fb, d: ['eng']).to_a
```

| selector | salaries returned | |
| --- | --- | --- |
| literal `"eng"` | `[200]` | correct |
| `$dept`, bound to `['eng']` | `[50, 200]` | the ops fact comes back too |
| `$d`, bound to `['eng']` | `[]` | nothing comes back |
| `$x`, bound to `['eng']` | `[]` | |

Outside `agg` the same parameters behave properly: `(eq dept $dept)` and `(eq dept $d)` both return the two eng facts.

The three answers differ only in the name of the parameter. `dept` is also a property, so `$dept` resolves per fact — for the ops fact the selector becomes "dept = ops", the aggregate is 50, and the ops fact matches its own maximum. `$d` and `$x` are not properties, resolve to nothing, and the selector matches nothing at all.

Note that the wrong answers are wrong quietly. No exception, no warning, just a different set of facts than the README's sentence describes.

## What should happen

Passing `fact` there looks deliberate, since it is what makes a correlated sub-query possible — the selector can refer to the fact being tested. If that is the intent, then the caller's parameters have to travel alongside it rather than instead of it, so both `$dept` from `each` and a property reference resolve as the reader expects, with a stated precedence between them.

If instead `agg` is meant to take parameters the way every other term does, the fix is to thread the params of the outer query into the call:

```ruby
def evaluate(fact, maps, fb, params = {})
  ...
  term.evaluate(nil, fb.query(selector, maps).each(fb, params).to_a, fb)
end
```

Either way the README example needs to match what the code does. As it stands it is the first thing a reader tries, and it returns facts that are not the ones the sentence promises.

## Related

#745 reports that both `agg` examples in the README use `(avg ...)`, which is not a term. This is the same two examples, failing for a second reason: even with a term that exists, the parameter does not work there.

gemshrine:
Fixes #798.

`Agg#evaluate` passed the current fact as the second argument of `Query#each`.
That argument is the query-parameter map. Therefore `$dept` inside an `agg`
selector resolved against the current fact's `dept` property and ignored the
`dept: ['eng']` supplied by the caller. The README example could return the
`ops` fact or no facts depending only on the parameter's name.

The selector now receives a context that gives explicit outer-query
parameters precedence, while falling back to the current fact for the existing
correlated-query behavior when no parameter with that name exists. The
aggregation API and literal selectors remain unchanged.

The regression builds two engineering salaries and one operations salary, then
runs the README-shaped query with `$dept`, `$d`, and `$x`; all must select only
the 200 salary. Each name produced a different wrong result on master.

Tests:

- `bundle exec rubocop lib/factbase/terms/agg.rb test/factbase/terms/test_agg.rb` — passed.

### DIFF
--- a/lib/factbase/terms/agg.rb
+++ b/lib/factbase/terms/agg.rb
@@ -29,6 +29,39 @@ class Factbase::Agg < Factbase::TermBase
     unless term.is_a?(Factbase::Term) || term.is_a?(Factbase::TermBase)
       raise(ArgumentError, "A term is expected, but '#{term}' provided")
     end
-    term.evaluate(nil, fb.query(selector, maps).each(fb, fact).to_a, fb)
+    term.evaluate(nil, fb.query(selector, maps).each(fb, params(fact)).to_a, fb)
+  end
+
+  private
+
+  # Extract the parameters made available to the outer query.
+  # @param [Factbase::Fact] fact The fact being evaluated
+  # @return [Hash] Parameters indexed by their names
+  def params(fact)
+    Context.new(fact, fact.all_properties.to_h { |name| [name, fact["$#{name}"]] }.compact)
+  end
+
+  # Values available to the selector of an aggregation.
+  class Context
+    # Ctor.
+    # @param [Factbase::Fact] fact The outer fact
+    # @param [Hash] params Parameters of the outer query
+    def initialize(fact, params)
+      @fact = fact
+      @params = params
+    end
+
+    # Get a parameter, or the property of the outer fact when it is not a parameter.
+    # @param [String] name Parameter or property name
+    # @return [Object] The value
+    def [](name)
+      @params.fetch(name) { @fact[name] }
+    end
+
+    # List all available names.
+    # @return [Array<String>] Names of the parameters and fact properties
+    def all_properties
+      @fact.all_properties | @params.keys
+    end
   end
 end
--- a/test/factbase/terms/test_agg.rb
+++ b/test/factbase/terms/test_agg.rb
@@ -71,4 +71,20 @@ class TestAgg < Factbase::Test
       )
     )
   end
+
+  def test_passes_outer_query_params_to_selector
+    fb = Factbase.new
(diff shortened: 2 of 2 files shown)


yegor256:
Merged, thanks! Good catch that the second argument of `Query#each` is the parameter map, not the fact.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1908](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #812, opened 2026-09-07
#796: Distinguish Boolean types in XML output

### LINKED ISSUE #796: ToXML marks a boolean with the same t attribute as Time and Float, since the type is the first letter of the class name
## What happens

`Factbase::ToXML` marks the type of every value with a one-letter `t` attribute, taken as the first character of the Ruby class name:

```ruby
def type_of(val)
  val.class.to_s[0]
end
```

Since 5a7bee5 (#748) a fact can hold a boolean, and `TrueClass` and `FalseClass` start with the same letters as `Time` and `Float`. The marker no longer identifies the type.

## Reproduction

```ruby
require 'factbase'
require 'factbase/to_xml'

fb = Factbase.new
f = fb.insert
f.tt = true
f.ff = false
f.fl = 2.5
f.tm = Time.parse('2026-01-01 10:00:00 UTC')
puts(Factbase::ToXML.new(fb).xml)
```

```xml
<f>
  <ff t="F">false</ff>
  <fl t="F">2.5</fl>
  <tm t="T">2026-01-01T10:00:00.000000Z</tm>
  <tt t="T">true</tt>
</f>
```

| value | class | `t` | collides with |
| --- | --- | --- | --- |
| `true` | `TrueClass` | `T` | `Time` |
| `false` | `FalseClass` | `F` | `Float` |
| `2.5` | `Float` | `F` | |
| the timestamp | `Time` | `T` | |

A reader of this document cannot tell `t="T"` meaning "a moment in time" from `t="T"` meaning "the boolean true", nor `t="F"` meaning "a float" from `t="F"` meaning "false". The only thing that separates them is guessing from the text content, which is exactly what the attribute exists to avoid.

`t="B"` is already spoken for as well: `put` uses it for a String that had to go out in Base64, so a class beginning with B would collide there too.

## Suggested fix

Spell the type instead of abbreviating the class name, so the mapping is a decision rather than an accident:

```ruby
TYPES = {
  'String' => 'S', 'Integer' => 'I', 'Float' => 'F', 'Time' => 'T',
  'TrueClass' => 'L', 'FalseClass' => 'L'
}.freeze

def type_of(val)
  TYPES.fetch(val.class.to_s) { raise("Cannot put #{val.class} into XML") }
end
```

with one letter for both booleans, since the text content already carries which one it is. Raising on an unknown class is worth having: the current code silently invents a letter for any type that ever gets added, which is how this collision arrived in the first place.

`Factbase::ToJSON` and `Factbase::ToYAML` do not have the problem, since both formats carry a native boolean.

gemshrine:
Fixes #796.

`ToXML` derived its one-letter type marker from the first character of the
Ruby class name. After Boolean values were accepted, `TrueClass` collided with
`Time` (`T`) and `FalseClass` with `Float` (`F`), so the attribute no longer
identified the value type.

The converter now uses an explicit mapping: String, Integer, Float, and Time
keep `S`, `I`, `F`, and `T`, while both Boolean classes use `L`. Unsupported
classes raise instead of silently introducing another accidental collision.

The regression serializes true, false, a Float, and a Time and checks each XML
attribute. Existing string, integer, float, time, and Base64-marker tests stay
unchanged.

Tests:

- `bundle exec rubocop lib/factbase/to_xml.rb test/factbase/test_to_xml.rb` — passed.

### DIFF
--- a/lib/factbase/to_xml.rb
+++ b/lib/factbase/to_xml.rb
@@ -25,6 +25,14 @@ require_relative '../factbase/flatten'
 # License:: MIT
 class Factbase::ToXML
   BAD = /[^\u0009\u000A\u000D\u0020-\uD7FF\uE000-\uFFFD\u{10000}-\u{10FFFF}]/
+  TYPES = {
+    'String' => 'S',
+    'Integer' => 'I',
+    'Float' => 'F',
+    'Time' => 'T',
+    'TrueClass' => 'L',
+    'FalseClass' => 'L'
+  }.freeze
 
   # Constructor.
   def initialize(fb, sorter = '_id')
@@ -80,6 +88,6 @@ class Factbase::ToXML
   end
 
   def type_of(val)
-    val.class.to_s[0]
+    TYPES.fetch(val.class.to_s) { raise(ArgumentError, "Can't put #{val.class} into XML") }
   end
 end
--- a/test/factbase/test_to_xml.rb
+++ b/test/factbase/test_to_xml.rb
@@ -94,6 +94,19 @@ class TestToXML < Factbase::Test
     ].each { |x| refute_empty(xml.xpath(x), out) }
   end
 
+  def test_marks_booleans_differently_from_floats_and_times
+    fb = Factbase.new
+    f = fb.insert
+    f.truth = true
+    f.lie = false
+    f.float = 2.5
+    f.time = Time.now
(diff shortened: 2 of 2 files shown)


yegor256:
Green, but this changes the `t` attribute contract of the XML export and starts raising on classes that were previously serialized without complaint. Anything already parsing that attribute has to learn `L`, so it wants a maintainer's call, and it is coupled to #813. Assigning to @yegor256.

---

PR #813, opened 2026-09-07
#795: Support Boolean literals in queries

### LINKED ISSUE #795: A boolean can be stored but not matched by any query, and type reports TrueClass for it
## What happens

5a7bee5 (#748) let a fact hold a boolean. The writer accepts it and `export`/`import` round-trips it, but the query language was not extended to match one, and the documentation still describes the older set of types.

## Reproduction

```ruby
require 'factbase'
fb = Factbase.new
f = fb.insert
f.b = true
f.c = false
f.n = 1
fb.query('(always)').each.to_a.first     # => [ b: [true], c: [false], n: [1] ]
```

| query | result | |
| --- | --- | --- |
| `(exists b)` | 1 | the property is there |
| `(eq b true)` | **0** | no way to match it |
| `(eq b "true")` | 0 | |
| `(eq b 1)` | 0 | |
| `(not b)` | 0 | |
| `(zero c)` | 0 | |
| `(as r (type b))` | `"TrueClass"` | |
| `(as r (type c))` | `"FalseClass"` | |
| `(as r (to_string b))` | `"true"` | |

A boolean can be written and read back through `Fact#[]`, and it survives `export`/`import`, but nothing in the query language selects on it. `(as r true)` evaluates to `nil`, so `true` is not a literal the parser knows — it is read as a property name, and the comparison quietly finds nothing.

## Two documentation gaps that came with it

`README.md:261` still says the factbase has

> no tables, schemas, or type enforcement beyond four scalar types: `Integer`, `Float`, `String`, and `Time`

There are six now.

And `(type v)` is documented as returning one of `"String"`, `"Integer"`, `"Float"`, `"Time"`, or `"Array"`. For a boolean it returns `"TrueClass"` or `"FalseClass"` — outside the documented set, and a Ruby class name rather than the plain word the other five are.

## Suggested fix

Three separate small things, in order of how much they cost a user:

1. Teach the tokenizer the `true` and `false` literals so `(eq b true)` works. Without it a stored boolean is write-only, which is a strange shape for a property the writer accepts.
2. Make `(type v)` answer `"Boolean"` for both, so it stays inside the vocabulary the README defines.
3. Update `README.md:261` to say six types, and add the boolean to the `(type v)` list.

A test that writes a boolean and then selects it by value would hold all three together.

gemshrine:
Fixes #795.

Facts already accepted `true` and `false`, and `export`/`import` preserved
them. The query parser did not recognize the same values as literals: it read
`true` as the property name `:true`. Consequently `(exists enabled)` found a
stored Boolean but `(eq enabled true)` quietly matched nothing, making a
writable fact value impossible to query by value.

`true` and `false` are now parsed before generic symbols. `(type value)` also
returns the public vocabulary `Boolean` for both Ruby Boolean classes instead
of leaking `TrueClass` or `FalseClass`. The README and API documentation list
the now-supported scalar values and result names.

The regression tests evaluate `(eq flag true)` and `(eq flag false)` against
stored facts and assert `Boolean` for each value. Both fail on master.

Tests:

- `bundle exec rubocop lib/factbase/syntax.rb lib/factbase/terms/type.rb lib/factbase.rb test/factbase/test_syntax.rb test/factbase/terms/test_type.rb` — passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ There are a few terms that return non-boolean values:
 * `(at i v)` is the `i`-th value of `v`
 * `(size v)` is the cardinality of `v` (zero if `v` is `nil`)
 * `(type v)` is the type of `v`
-(`"String"`, `"Integer"`, `"Float"`, `"Time"`, or `"Array"`)
+(`"String"`, `"Integer"`, `"Float"`, `"Time"`, `"Boolean"`, or `"Array"`)
 * `(either v1 v2)` is `v2` if `v1` is `nil`
 
 It's possible to modify the facts retrieved, on fly:
@@ -259,8 +259,8 @@ There are some system-level terms:
 
 The entire database is a single flat [Ruby](https://www.ruby-lang.org/en/)
   `Array` of `Hash` objects held in RAM (`Factbase#@maps`). There are no
-  tables, schemas, or type enforcement beyond four scalar types: `Integer`,
-  `Float`, `String`, and `Time`. This contrasts with
+  tables, schemas, or type enforcement beyond six scalar types: `Integer`,
+  `Float`, `String`, `Time`, `TrueClass`, and `FalseClass`. This contrasts with
   [SQLite](https://sqlite.org/) (fixed-column tables on disk) and
   [MongoDB](https://www.mongodb.com/) (typed document collections). New
   programmers must understand that all data vanishes on process exit unless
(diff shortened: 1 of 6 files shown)


yegor256:
Green and well tested, but this widens the query language: `true`/`false` become reserved literals, so a fact with a property literally named `true` can no longer be referenced by name, and `(type v)` starts answering a new word, `Boolean`. That is a language decision rather than a defect fix, and it is coupled to #812, which changes the `t` markers in the XML output. Both want a maintainer's judgement together. Assigning to @yegor256.

---

PR #814, opened 2026-09-07
#794: Document zero-based nth positions

### LINKED ISSUE #794: The nth term is zero-based and accepts zero, while the README calls for a positive one-based position
## What happens

The README says:

> * `(nth v p)` returns the `p` property of the _v_-th fact (must be a positive integer)

Two things in one line disagree with the code. Zero is accepted, so the argument is not required to be positive. And the indexing is zero-based, so the "v-th fact" is not the one the reader counts to.

## Reproduction

```ruby
require 'factbase'
fb = Factbase.new
[1, 5, 3, 9].each { |n| fb.insert.n = n }
fb.query('(as r (agg (always) (nth 0 n)))').each.to_a.map { |f| f['r'] }  # => [[1]]
fb.query('(as r (agg (always) (nth 1 n)))').each.to_a.map { |f| f['r'] }  # => [[5]]
fb.query('(as r (agg (always) (first n)))').each.to_a.map { |f| f['r'] }  # => [[1]]
```

| query | result | |
| --- | --- | --- |
| `(nth 0 n)` | 1 | accepted, though the README calls for a positive integer |
| `(nth 1 n)` | 5 | the second fact, not the first |
| `(first n)` | 1 | the first fact |
| `(nth -1 n)` | `A non-negative position is expected, but -1 provided` | the message says non-negative, the README says positive |
| `(nth 99 n)` | `nil` | out of range answers quietly |

So `(nth 1 p)` and `(first p)` name different facts, which is the trap this line of the README walks the reader straight into: someone reading "the v-th fact, must be a positive integer" writes `(nth 1 n)` for the first fact and silently gets the second.

The error message from the term itself contradicts the README as well: it demands a non-negative position, which admits the zero the README forbids.

## Suggested fix

The cheapest fix is the README, since the behaviour is a plain array index and the error message already describes it correctly:

```
* `(nth v p)` returns the `p` property of the fact at position `v`, counting
  from zero (must be a non-negative integer); `(nth 0 p)` is the same as `(first p)`
```

Making the term one-based instead would match the current wording, but it would change the meaning of every existing `(nth ...)` in the wild, so the README looks like the side that should move.

gemshrine:
Fixes #794.

`nth` has always used an array position: it accepts zero, rejects only negative
values, and `(nth 0 p)` returns the same first value as `(first p)`. The README
said the opposite — "the v-th fact" with a positive integer — so a reader
naturally wrote `(nth 1 p)` for the first fact and silently received the
second one.

Changing the term to one-based indexing would alter every existing query. The
implementation and its validation message already agree with each other, so
this change moves the documentation to the stable, zero-based contract instead:
`v` is a non-negative position and `(nth 0 p)` is `(first p)`.

The aggregation tests already exercise `(nth 0 x)` and establish the existing
behavior. No runtime code is changed.

Tests:

- `bundle exec rake test` — passed, 98.84% coverage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,8 +231,9 @@ with the salary average in their departments
 There are also terms that match the entire factbase
 and must be used primarily inside the `(agg ..)` term:
 
-* `(nth v p)` returns the `p` property of the _v_-th fact (must be
-a positive integer)
+* `(nth v p)` returns the `p` property of the fact at position `v`, counting
+  from zero (must be a non-negative integer); `(nth 0 p)` is the same as
+  `(first p)`
 * `(first p)` returns the `p` property of the first fact
 * `(count)` returns the tally of facts
 * `(max p)` returns the maximum value of the `p` property in all facts


yegor256:
Merged, thanks! Documenting the stable zero-based contract is the right call here, rather than moving every existing query.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (5) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1884](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #815, opened 2026-09-07
#793: Check rules before writing fact properties

### LINKED ISSUE #793: Factbase::Rules outside a transaction keeps the fact it refuses, and lets an unassigned fact through unchecked
## What happens

`Factbase::Rules` is documented as the way to make sure a fact cannot enter the factbase unless it matches a rule. Inside a transaction it works: the check runs at commit and a violation rolls the whole transaction back. Outside a transaction it does not hold at all. The rule is applied once per assignment rather than once per fact, nothing is ever undone, and a fact that never gets a single assignment is never checked.

## Reproduction

```ruby
require 'factbase'
require 'factbase/rules'

fb = Factbase::Rules.new(Factbase.new, '(exists foo)')
fb.insert                                   # no assignment at all
fb.query('(always)').each.to_a.size         # => 1, and the fact has no foo
```

Four cases, all with the rule `(exists foo)`:

| what the caller does | result | factbase afterwards |
| --- | --- | --- |
| `fb.insert` and nothing else | no error | holds `[ ]`, which violates the rule |
| `f.bar = 3` | `ArgumentError` | holds `[ bar: [3] ]`, the value the rule rejected |
| `f.foo = 1` then `f.bar = 3` | no error | holds `[ bar: [3], foo: [1] ]` |
| `f.bar = 3` then `f.foo = 1` | `ArgumentError` on the first line | holds `[ bar: [3] ]` |

The last two are the same fact built in a different order. One is accepted, the other is rejected and then left behind half-written. Anything that fills a fact from a hash is at the mercy of the hash ordering.

The second row is the worse one for a caller who rescues: the error says the fact was refused, and the refused fact is in the factbase.

## Root cause

`lib/factbase/rules.rb`, the `Fact` decorator:

```ruby
others do |*args|
  r = @fact.method_missing(*args)
  k = args.first.to_s
  @check.it(@fact, @fb) if k.end_with?('=')
  r
end
```

The write to the underlying fact happens on the first line, the check on the third. Outside a transaction there is no journal, so raising after the write cannot take it back. And because the check is bound to the assignment, a fact created by `insert` and never assigned to is never presented to `Check#it` at all.

Inside `txn` none of this bites, because `@check` is swapped for `Later`, which stays quiet, and the real check runs over the committed facts in `Factbase::Rules#txn`.

## Suggested fix

Check before writing, so a violation leaves nothing behind:

```ruby
others do |*args|
  k = args.first.to_s
  unless k.end_with?('=')
    return @fact.method_missing(*args)
  end
  probe = Factbase::Accum.new(@fact, {}, false)
  probe.method_missing(*args)
  @check.it(probe, @fb)
  @fact.method_missing(*args)
end
```

`Factbase::Accum` already exists for exactly this, holding the new property without passing it to the fact when `pass` is false.

That still leaves the per-assignment semantics, under which `f.bar = 3` before `f.foo = 1` is refused. If the intent is that a fact is judged once it is complete, then the non-transactional path needs the same deferral the transactional one has, and `insert` outside a transaction should be checked too rather than only on assignment.

gemshrine:
Fixes #793.

`Factbase::Rules::Fact` applied an assignment to its wrapped fact before it
checked the rule. Outside `txn`, there is no journal to undo that write. With
the rule `(exists foo)`, `fact.bar = 42` raised as intended but left
`bar: [42]` in the factbase — a caller rescuing the error was told that the
fact was rejected while the rejected value remained stored.

The order now matches `Factbase::Inv`: a temporary `Factbase::Accum` receives
the pending assignment with `pass: false`, the rule is evaluated against that
candidate, and only a successful candidate is forwarded to the actual fact.
Reads and assignments that satisfy the rule keep their previous path.

This deliberately does not change when a newly inserted, still-empty fact is
considered complete. There is no completion event outside a transaction, and
turning `insert` itself into a rule check would be a separate contract change.
The PR fixes the concrete broken atomicity of an assignment.

The regression test creates a Rules-wrapped factbase with `(exists foo)`,
attempts `bar = 42`, asserts the error, and then reads the wrapped factbase to
prove that `bar` was never written. It fails on master.

Tests:

- `bundle exec rubocop lib/factbase/rules.rb test/factbase/test_rules.rb` — passed.

### DIFF
--- a/lib/factbase/rules.rb
+++ b/lib/factbase/rules.rb
@@ -6,6 +6,7 @@
 require 'decoor'
 require 'others'
 require_relative '../factbase'
+require_relative '../factbase/accum'
 require_relative '../factbase/syntax'
 require_relative '../factbase/tallied'
 
@@ -84,10 +85,13 @@ class Factbase::Rules
     end
 
     others do |*args|
-      r = @fact.method_missing(*args)
       k = args.first.to_s
-      @check.it(@fact, @fb) if k.end_with?('=')
-      r
+      if k.end_with?('=')
+        probe = Factbase::Accum.new(@fact, {}, false)
+        probe.method_missing(*args)
+        @check.it(probe, @fb)
+      end
+      @fact.method_missing(*args)
     end
   end
 
--- a/test/factbase/test_rules.rb
+++ b/test/factbase/test_rules.rb
@@ -152,6 +152,13 @@ class TestRules < Factbase::Test
     refute_includes(ex.message, 'nil', 'Error message should not contain "nil"')
   end
 
+  def test_does_not_write_a_value_rejected_by_a_rule
+    base = Factbase.new
+    fact = Factbase::Rules.new(base, '(exists foo)').insert
+    assert_raises(ArgumentError) { fact.bar = 42 }
+    assert_nil(base.query('(always)').each.to_a.first['bar'])
+  end
+
   def test_error_message_truncates_long_expression
     rule = '(and (exists a) (exists b) (exists c) (exists d) (exists e))'
     assert_operator(rule.length, :>, 32)


yegor256:
Merged, thanks! Probing through `Accum` with `pass: false` keeps the existing properties visible to the rule while the pending write stays out of the fact, which is exactly what was needed.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1084](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #816, opened 2026-09-07
#792: Name query terms in arity errors

### LINKED ISSUE #792: Twenty-eight terms lose their name in the too-few-operands error, reporting an empty string or a Ruby method
## What happens

When a term is called with too few operands, the error is supposed to name the term. For eleven terms it does. For twenty-eight it does not: the name comes out either empty or as the Ruby method the term happens to be built on, which the user never typed and cannot find in the README.

```ruby
require 'factbase'
fb = Factbase.new
fb.insert.foo = 1
fb.query('(size)').each.to_a
```

| written by the user | message says | |
| --- | --- | --- |
| `(size)` | `Too few (0) operands for '' (1 expected)` | name lost |
| `(type)` | `Too few (0) operands for '' (1 expected)` | name lost |
| `(contains)` | `Too few (0) operands for 'include?' (2 expected)` | Ruby method |
| `(starts_with)` | `Too few (0) operands for 'start_with?' (2 expected)` | Ruby method |
| `(gte)` | `Too few (0) operands for '>=' (2 expected)` | Ruby operator |
| `(plus)` | `Too few (0) operands for '+' (2 expected)` | Ruby operator |
| `(either)` | `Too few (0) operands for 'either' (2 expected)` | correct |
| `(agg)` | `Too few (0) operands for 'agg' (2 expected)` | correct |

The full list of terms that lose their name: `nil`, `exists`, `absent`, `zero`, `eq`, `lt`, `gt`, `lte`, `gte`, `many`, `one`, `matches`, `contains`, `starts_with`, `ends_with`, `size`, `type`, `as`, `join`, `plus`, `minus`, `times`, `div`, `to_integer`, `to_string`, `to_float`, `defn`, `undef`.

The ones that get it right: `at`, `either`, `unique`, `agg`, `sum`, `min`, `max`, `first`, `sorted`, `head`, `inverted`, `assert`.

An empty name is the worse half. `Too few (0) operands for ''` tells the reader nothing at all, and a query holding several terms gives no clue which one is short of operands. The `(size)` at the end of the message is the only hint, and it is there by accident, from `#to_s`.

## Root cause

`assert_args` reports `@op`, and the classes that back these terms either never set `@op` or set it to the Ruby method they delegate to. `Factbase::Compare#initialize` takes the Ruby operator as its `@op` (`:>=`, `:include?`, `:start_with?`), `Factbase::Arithmetic#initialize` takes `:+`, `:-`, `:*`, `:/`, and the single-operand classes such as `Factbase::Size` and `Factbase::Type` leave it nil.

The terms that report correctly are the ones whose class sets `@op` to the name from the query.

## Suggested fix

Pass the term name as written down to the class, and keep the Ruby method separate from it. `Factbase::Term` knows the name it dispatched on, so the simplest fix is for `assert_args` to use that name rather than whatever the class stored:

```ruby
def assert_args(num)
  raise(ArgumentError, "Too few (#{@operands.size}) operands for '#{name}' (#{num} expected)") if @operands.size < num
  ...
end
```

with `name` falling back to `@op` for the classes that already set it properly.

gemshrine:
Fixes #792.

`assert_args` reported the implementation's `@op`. That was not consistently
the query term: simple terms left it empty, comparison terms stored operators
such as `include?` or `>=`, and arithmetic terms stored `+`, `-`, `*`, or
`/`. Thus `(size)` became `Too few ... for ''`, while `(contains)` named a Ruby
method that a reader never wrote and cannot find in the README.

`Factbase::Term` is the single place that knows the operator parsed from the
query. It now gives that name to every dispatch target when it builds its term
table. A target may be a thin wrapper around another `TermBase` implementation
(for example, `contains` wraps a comparison), so the name travels to those
helpers as well. Arithmetic and comparison execution still use their existing
Ruby operators; only diagnostics have a separate, user-facing name.

The regression test evaluates zero-operand `size`, `contains`, `gte`, and
`plus` against a fact. They cover an unset `@op`, a Ruby method, a Ruby
operator, and an arithmetic wrapper. Each message now contains exactly the
term the query used; all four fail on master.

Tests:

- `bundle exec rubocop lib/factbase/terms/base.rb lib/factbase/term.rb test/factbase/test_syntax.rb` — passed.

### DIFF
--- a/lib/factbase/term.rb
+++ b/lib/factbase/term.rb
@@ -156,7 +156,7 @@ class Factbase::Term < Factbase::TermBase
     super()
     @op = operator
     @operands = operands
-    @terms = TERMS.transform_values { |c| c.new(operands) }
+    @terms = TERMS.to_h { |name, klass| [name, klass.new(operands).tap { |term| term.name = name }] }
   end
 
   # Extend it with the module.
--- a/lib/factbase/terms/base.rb
+++ b/lib/factbase/terms/base.rb
@@ -8,6 +8,18 @@
 # Copyright:: Copyright (c) 2024-2026 Yegor Bugayenko
 # License:: MIT
 class Factbase::TermBase
+  # Set the name the user used for this term and its implementation helpers.
+  # @param [Symbol] name Name of the term in the query
+  # rubocop:disable Elegant/GoodMethodName
+  def name=(name)
+    @name = name
+    instance_variables.each do |variable|
+      value = instance_variable_get(variable)
+      value.name = name if value.is_a?(Factbase::TermBase)
+    end
+  end
+  # rubocop:enable Elegant/GoodMethodName
+
   # Turns it into a string.
   # @return [String] The string of it
   def to_s
@@ -33,8 +45,9 @@ class Factbase::TermBase
 
   def assert_args(num)
     c = @operands.size
-    raise(ArgumentError, "Too many (#{c}) operands for '#{@op}' (#{num} expected)") if c > num
-    raise(ArgumentError, "Too few (#{c}) operands for '#{@op}' (#{num} expected)") if c < num
+    name = @name || @op
+    raise(ArgumentError, "Too many (#{c}) operands for '#{name}' (#{num} expected)") if c > num
+    raise(ArgumentError, "Too few (#{c}) operands for '#{name}' (#{num} expected)") if c < num
   end
 
   # Turns facts into plain maps, keeping only the properties they really carry.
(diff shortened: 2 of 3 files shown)


yegor256:
Merged, thanks! Naming the query term instead of the Ruby operator makes these errors findable in the README.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1864](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #817, opened 2026-09-07
#791: Explain invalid substring comparison operands

### LINKED ISSUE #791: A non-string left operand makes contains, starts_with and ends_with report themselves as not defined
## What happens

`Factbase::Term#evaluate` treats every `NoMethodError` raised anywhere below it as proof that the term does not exist. `lib/factbase/term.rb:208`:

```ruby
rescue NoMethodError => e
  raise(RuntimeError, "Probably the term '#{@op}' is not defined at #{self}: #{e.message}")
```

The three substring terms are built on `Factbase::Compare`, which sends `include?`, `start_with?` or `end_with?` to the left value. When the left value is not a String, Ruby raises `NoMethodError`, the rescue swallows it, and the user is told the term is not defined — for a term the README documents.

## How to reproduce

```ruby
require 'factbase'
fb = Factbase.new
f = fb.insert
f.s = 'hello'
f.n = 7
fb.query('(contains n 7)').each.to_a
```

| query | message |
| --- | --- |
| `(contains n 7)` | `Probably the term 'contains' is not defined at (contains n 7): undefined method 'include?' for an instance of Integer` |
| `(starts_with n 7)` | `Probably the term 'starts_with' is not defined ...: undefined method 'start_with?'` |
| `(ends_with n 7)` | `Probably the term 'ends_with' is not defined ...: undefined method 'end_with?'` |
| `(contains s 7)` | `"no implicit conversion of Integer into String" at (contains s 7)` |

The last row is the same mistake made on the other operand, and there the message is accurate and useful. Only the left operand produces the false claim, and the claim is confidently wrong: `contains` is defined, it is in the README, and the reader is sent looking for a missing term instead of a numeric property.

This is the shape of #742, which was accepted and fixed for `when` alone. The rescue that produces it is still there for everything else.

## Suggested fix

Let `Factbase::Compare` say what actually went wrong, so the rescue never sees the `NoMethodError`:

```ruby
def _compare(left, right)
  unless left.respond_to?(@op)
    raise(
      ArgumentError,
      "Cannot apply (compare #{@op}) to #{left.inspect} (#{left.class}), it is not a String"
    )
  end
  left.__send__(@op, right)
rescue ArgumentError => e
  ...
end
```

Narrowing the rescue in `term.rb:208` to the case where `@op` really is unknown would fix the whole class of misreports rather than these three terms, since any `NoMethodError` from any operand currently comes out as "the term is not defined".

gemshrine:
Fixes #791.

`Factbase::Term#evaluate` reserves its `NoMethodError` rescue for a term that
does not exist. `contains`, `starts_with`, and `ends_with` are implemented by
`Factbase::Compare`, however, and send `include?`, `start_with?`, or
`end_with?` to the left value. On an Integer that send raised `NoMethodError`,
which escaped to the outer rescue and became the false statement that the
documented query term was not defined.

`Compare` now handles that error alongside its existing `ArgumentError`
handling. The message keeps both values, their classes, the comparison
operation, and Ruby's original explanation. The outer term-level rescue is
unchanged, so an actually unknown query term still receives its existing
diagnostic.

The regression constructs `Compare` with `include?` and two Integers. It
asserts the message names the values and `(compare include?)`, as well as the
original missing-method explanation. It fails on master, where the inner
`NoMethodError` was allowed to be misclassified by `Term#evaluate`.

Tests:

- `bundle exec rubocop lib/factbase/terms/compare.rb test/factbase/terms/test_compare.rb` — passed.

### DIFF
--- a/lib/factbase/terms/compare.rb
+++ b/lib/factbase/terms/compare.rb
@@ -44,7 +44,7 @@ class Factbase::Compare < Factbase::TermBase
   # @return [Boolean] The result of the comparison
   def _compare(left, right)
     left.__send__(@op, right)
-  rescue ArgumentError => e
+  rescue ArgumentError, NoMethodError => e
     raise(
       RuntimeError,
       "Cannot compare #{left.inspect} (#{left.class}) " \
--- a/test/factbase/terms/test_compare.rb
+++ b/test/factbase/terms/test_compare.rb
@@ -42,4 +42,12 @@ class TestCompare < Factbase::Test
     assert_includes(e.message, '"yesterday" (String)')
     assert_includes(e.message, 'comparison of Time with String failed')
   end
+
+  def test_wraps_a_missing_comparison_method
+    t = Factbase::Compare.new(:include?, [42, 7])
+    e = assert_raises(RuntimeError) { t.evaluate(fact, [], Factbase.new) }
+    assert_includes(e.message, 'Cannot compare 42 (Integer) with 7 (Integer)')
+    assert_includes(e.message, 'using (compare include?)')
+    assert_includes(e.message, "undefined method 'include?'")
+  end
 end


yegor256:
Merged, thanks! The message no longer claims a documented term does not exist.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1064](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #818, opened 2026-09-07
#790: Aggregate multi-valued arithmetic operands

### LINKED ISSUE #790: Arithmetic terms refuse a multi-valued operand, while the README promises they sum and multiply all values
## What happens

The README describes the four arithmetic terms as aggregating over every value of an operand:

> * `(plus v1 v2)` is a sum of `∑v1` and `∑v2`
> * `(minus v1 v2)` is a deduction of `∑v2` from `∑v1`
> * `(times v1 v2)` is a multiplication of `∏v1` and `∏v2`
> * `(div v1 v2)` is a division of `∏v1` by `∏v2`

The sigma and pi are the whole point of that wording: a property in a factbase normally holds an array, and the README says so a few lines above — "each assignment appends... New programmers must expect multi-element arrays on every property read."

The implementation refuses any operand that holds more than one value.

## How to reproduce

```ruby
require 'factbase'
fb = Factbase.new
f = fb.insert
f.n = 7
f.x = 2
f.x = 3          # x is now [2, 3], the normal shape of a property
fb.query('(as r (plus x n))').each.to_a
```

| query | README | actual |
| --- | --- | --- |
| `(plus x n)` | `(2+3) + 7` = 12 | `RuntimeError: Too many values at first position, one expected` |
| `(minus x n)` | `(2+3) - 7` = -2 | same error |
| `(times x n)` | `(2*3) * 7` = 42 | same error |
| `(div x n)` | `(2*3) / 7` | same error |

Every single-valued case works, so the terms are usable only on properties that happen to hold exactly one value — which the README explicitly tells the reader not to assume.

## Root cause

`lib/factbase/terms/arithmetic.rb:28` and `:31`:

```ruby
lefts = _values(0, fact, maps, fb)
return if lefts.nil?
raise(ArgumentError, 'Too many values at first position, one expected') unless lefts.size == 1
rights = _values(1, fact, maps, fb)
return if rights.nil?
raise(ArgumentError, 'Too many values at second position, one expected') unless rights.size == 1
```

There is no aggregation anywhere in the class; `v` and `r` are simply `lefts[0]` and `rights[0]`.

## What should happen

Either the terms aggregate the way the README says:

```ruby
v = @op == :* || @op == :/ ? lefts.reduce(:*) : lefts.sum
r = @op == :* || @op == :/ ? rights.reduce(:*) : rights.sum
```

(with the `Time` plus `String` branch kept for the single-value case), or the README drops the sigma and pi and states that both operands must hold exactly one value.

The second one is a smaller change but a worse contract, since it leaves the most common shape of a property unusable in arithmetic. Whichever way it goes, the two should agree.

gemshrine:
Fixes #790.

The README defines `plus` and `minus` in terms of the sum of every value of
each operand, and `times` and `div` in terms of their products. Facts normally
store an array after repeated assignment, but `Arithmetic` rejected any array
larger than one before it performed an operation. Thus `x = 2; x = 3` made
`(plus x n)` fail rather than calculate `(2 + 3) + n`.

Each operand is now reduced before the binary operation: addition/subtraction
use a sum, multiplication/division a product. Single-valued operands retain
their exact existing value, which preserves the `Time` plus/minus duration
path and the existing zero-divisor check. The change does not invent an
aggregation for missing operands; they still return `nil`.

The regression test uses two multi-valued numeric properties, `[2, 3]` and
`[7, 2]`, and exercises all four operations. It expects 14, -4, 84, and 0
under Ruby integer division. Before this PR each case raised the old
"Too many values" error.

Tests:

- `bundle exec rubocop lib/factbase/terms/arithmetic.rb test/factbase/terms/test_math.rb` — passed.

### DIFF
--- a/lib/factbase/terms/arithmetic.rb
+++ b/lib/factbase/terms/arithmetic.rb
@@ -25,12 +25,10 @@ class Factbase::Arithmetic < Factbase::TermBase
     assert_args(2)
     lefts = _values(0, fact, maps, fb)
     return if lefts.nil?
-    raise(ArgumentError, 'Too many values at first position, one expected') unless lefts.size == 1
     rights = _values(1, fact, maps, fb)
     return if rights.nil?
-    raise(ArgumentError, 'Too many values at second position, one expected') unless rights.size == 1
-    v = lefts[0]
-    r = rights[0]
+    v = aggregate(lefts)
+    r = aggregate(rights)
     if v.is_a?(Time) && r.is_a?(String)
       (num, units) = r.split
       begin
@@ -56,5 +54,15 @@ class Factbase::Arithmetic < Factbase::TermBase
     end
     raise(ArgumentError, 'Cannot divide by zero') if @op == :/ && r.is_a?(Numeric) && r.zero?
     v.__send__(@op, r)
+  end
+
+  private
+
+  # Aggregate every value supplied for one arithmetic operand.
+  # @param [Array] values Values of one operand
+  # @return [Object] Sum or product of the values
+  def aggregate(values)
+    return values[0] if values.size == 1
+    values.reduce(@op == :+ || @op == :- ? :+ : :*)
   end
 end
--- a/test/factbase/terms/test_math.rb
+++ b/test/factbase/terms/test_math.rb
@@ -102,6 +102,18 @@ class TestMath < Factbase::Test
     assert_nil(t.evaluate(fact, [], Factbase.new))
   end
 
+  def test_aggregates_multiple_values_before_arithmetic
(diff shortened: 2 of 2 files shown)


yegor256:
Green, but this changes what arithmetic terms mean for multi-valued operands: a query that used to fail loudly now returns a number, and `div` reduces by product, so `(div x y)` over two-valued properties is neither of the two divisions a reader might expect. The README does promise this, so the direction looks right — I'd still rather a maintainer confirmed the aggregation choice per operator before it becomes the contract. Assigning to @yegor256.

---

PR #819, opened 2026-09-07
#784: Skip One indexing for subterm operands

### LINKED ISSUE #784: IndexedOne treats a sub-term operand as a property name, so (one (eq kind "a")) returns nothing
## What happens

`(one v)` is documented as "if `v` has one value", and `Factbase::One#evaluate` puts no restriction on what `v` is: it calls `_values(0, fact, maps, fb)`, which happily evaluates a nested term and yields its result as the single value. So `(one (eq kind "a"))` is true for every fact, because the nested term always produces exactly one value.

Under `Factbase::IndexedFactbase` the same query silently answers nothing. No exception, no warning, just an empty result where the plain factbase returns every fact.

## How to reproduce

```ruby
require 'factbase'

def build(fb)
  f = fb.insert; f.kind = 'a'; f.n = 1
  f = fb.insert; f.kind = 'b'; f.n = 2
  fb
end

build(Factbase.new).query('(one (eq kind "a"))').each.to_a.size                          # => 2
build(Factbase::IndexedFactbase.new(Factbase.new)).query('(one (eq kind "a"))').each.to_a.size  # => 0
```

| query | plain `Factbase` | `IndexedFactbase` |
| --- | --- | --- |
| `(one (eq kind "a"))` | 2 facts | **0 facts** |
| `(one (plus n 1))` | 2 facts | **0 facts** |
| `(one kind)` | 2 facts | 2 facts |
| `(many (eq kind "a"))` | 0 facts | 0 facts |

`(one kind)` agrees, because there the operand really is a property. `(many ...)` agrees by accident: there is no `IndexedMany`, so it falls back to the plain term.

## Root cause

`lib/factbase/indexed/indexed_one.rb:14-15`:

```ruby
prop = @term.operands.first.to_s
key = [maps.object_id, prop, @term.op]
```

and line 26:

```ruby
entry[:facts] << f if !f[prop].nil? && f[prop].size == 1
```

The operand is assumed to be a property name. When it is a term, `to_s` turns it into the string `"(eq kind \"a\")"`, `f[prop]` looks up a property under that literal name, finds nothing, and every fact is dropped.

The comparison terms guard against exactly this. `indexed_gte.rb:14` starts with

```ruby
return unless op1.is_a?(Symbol) && _scalar?(op2)
```

so a non-property operand makes them decline the optimization and let the plain term answer. `IndexedOne` has no such check.

## Suggested fix

Give `IndexedOne` the same guard the comparison terms have, so a non-property operand falls through to `Factbase::One`:

```ruby
def predict(maps, _fb, _params)
  op = @term.operands.first
  return unless op.is_a?(Symbol)
  prop = op.to_s
  ...
end
```

A test asserting that `(one <sub-term>)` returns the same facts with and without the decorator would keep the two paths together.

gemshrine:
Fixes #784.

`IndexedOne` assumed that the operand of `(one ...)` was always a property
name. For `(one (eq kind "a"))`, it converted the nested term to the literal
property name `(eq kind "a")`, found that no fact had such a property, and
returned an empty candidate list. The unindexed `One` evaluates subterms as a
single value, so the two factbase implementations gave different answers.

The index now explicitly accepts only a Symbol operand. This retains the fast
property-name index for `(one kind)` and declines optimization for a nested
term. `IndexedTerm` treats `nil` as "no safe prediction", so the standard
term evaluator handles the subterm and preserves plain Factbase semantics.

The regression redresses `(one (eq kind "a"))` with `IndexedTerm` and asserts
that prediction is declined. It fails on master, where the index returned an
empty array. Existing tests still cover Array, Taped, and LazyTaped property
maps, ensuring the optimized path remains intact.

Tests:

- `bundle exec rubocop lib/factbase/indexed/indexed_one.rb test/factbase/indexed/test_indexed_one.rb` — passed.

### DIFF
--- a/lib/factbase/indexed/indexed_one.rb
+++ b/lib/factbase/indexed/indexed_one.rb
@@ -11,7 +11,9 @@ class Factbase::IndexedOne
   end
 
   def predict(maps, _fb, _params)
-    prop = @term.operands.first.to_s
+    operand = @term.operands.first
+    return unless operand.is_a?(Symbol)
+    prop = operand.to_s
     key = [maps.object_id, prop, @term.op]
     @idx[key] ||= { facts: [], count: 0 }
     entry = @idx[key]
--- a/test/factbase/indexed/test_indexed_one.rb
+++ b/test/factbase/indexed/test_indexed_one.rb
@@ -40,6 +40,12 @@ class TestIndexedOne < Factbase::Test
     end
   end
 
+  def test_skips_a_subterm_operand
+    term = Factbase::Term.new(:one, [Factbase::Term.new(:eq, [:kind, 'a'])])
+    term.redress!(Factbase::IndexedTerm, idx: {})
+    assert_nil(term.predict([{ 'kind' => ['a'] }], nil, {}))
+  end
+
   private
 
   def _assert_one


yegor256:
Merged, thanks! Declining the prediction instead of guessing a property name is the right call — indexed and plain `Factbase` agree again.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1044](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #820, opened 2026-09-07
#783: Preserve range errors with mixed indexed values

### LINKED ISSUE #783: Indexed range terms raise a bare ArgumentError where plain Factbase raises a contextual RuntimeError
## What happens

`Factbase::IndexedFactbase` is described in the README as a performance decorator: without it "every `query#each` call performs a full linear scan". It also changes what a range query raises when a property holds values of more than one type.

`Factbase::Compare#_compare` (`lib/factbase/terms/compare.rb:45-53`) catches Ruby's `ArgumentError` and re-raises a `RuntimeError` that names the two values, their classes and the operator. The four indexed comparison terms compare raw values themselves and never reach that guard, so the caller gets a bare `ArgumentError` instead.

## How to reproduce

```ruby
require 'factbase'

def build(fb)
  f = fb.insert; f.kind = 'good'; f.n = 5
  f = fb.insert; f.kind = 'good'; f.n = 7
  f = fb.insert; f.kind = 'bad';  f.n = 'oops'
  fb
end

build(Factbase.new).query('(gt n 0)').each.to_a
build(Factbase::IndexedFactbase.new(Factbase.new)).query('(gt n 0)').each.to_a
```

| query | plain `Factbase` | `IndexedFactbase` |
| --- | --- | --- |
| `(gt n 0)` | `RuntimeError: Cannot compare "oops" (String) with 0 (Integer) using (compare >): comparison of String with 0 failed` | `ArgumentError: comparison of String with 7 failed` |
| `(gte n 1)` | `RuntimeError: Cannot compare "oops" (String) with 1 (Integer) using (compare >=): ...` | `ArgumentError: comparison of Integer with String failed` |
| `(lt n 100)` | `RuntimeError: Cannot compare "oops" (String) with 100 (Integer) using (compare <): ...` | `ArgumentError: comparison of Integer with String failed` |

Two things go wrong, not one. The exception class differs, so a caller that rescues the `RuntimeError` the plain factbase raises stops catching anything once the decorator is added. And the message names a value the query never mentioned: `(gt n 0)` reports a failure "with 7", because the comparison that actually blew up is the internal sort of the index, not the comparison the user asked for. There is nothing in it to say which property or which term is at fault.

## Root cause

`lib/factbase/indexed/indexed_gte.rb:40`:

```ruby
```

and line 45:

```ruby
idx = entry[:facts].bsearch_index { |v, _| v >= target }
```

Both compare values straight from the property, so `Array#sort_by!` raises the bare `ArgumentError` before any guard sees it. The same two lines are in `indexed_gt.rb:40,45`, `indexed_lt.rb:40,45` and `indexed_lte.rb:40,45`.

## Suggested fix

Wrap the two comparison sites the way `Compare#_compare` does, so the decorated factbase raises the same error as the undecorated one:

```ruby
def _feed(facts, entry, prop)
  return unless entry[:count] < facts.size
  facts[entry[:count]..].each do |fact|
    fact[prop]&.each { |v| entry[:facts] << [v, fact] }
  end
  begin
  rescue ArgumentError => e
    raise(RuntimeError, "Cannot index #{prop.inspect} for (compare #{@term.op}), its values have different types: #{e.message}")
  end
  entry[:count] = facts.size
end
```

Alternatively the four terms could give up and return `maps` untouched when the sort fails, letting the plain term run and produce the contextual error by itself.

Worth a test that runs the same query against `Factbase` and `Factbase::IndexedFactbase` and asserts the same exception class, since the decorator is meant to be invisible.

gemshrine:
Fixes #783.

The range indexes sort raw property values before they select candidates. When
a property held both an Integer and a String, that internal sort raised a bare
`ArgumentError`. Plain Factbase does not sort first: `Compare` evaluates the
requested operator and raises a contextual `RuntimeError` that identifies the
two values and their types. Adding `IndexedFactbase` therefore changed both
the error class and the explanation of the same query.

Each indexed range term now checks whether the candidate property values can
be sorted before it creates or extends its cache. If they cannot, prediction
returns `nil` and the ordinary comparison path runs. No partial index entry is
stored, so later queries do not inherit a failed sort. Homogeneous properties
retain the existing index and binary-search path.

The regression runs `gt`, `gte`, `lt`, and `lte` against the same mixed
Integer/String facts through plain and IndexedFactbase. For every operation it
asserts both raise `RuntimeError` with identical messages. The indexed path
raised a bare `ArgumentError` on master.

Tests:

- `bundle exec rake test` — passed, 98.86% coverage;
- `bundle exec rubocop lib/factbase/indexed/indexed_gt.rb lib/factbase/indexed/indexed_gte.rb lib/factbase/indexed/indexed_lt.rb lib/factbase/indexed/indexed_lte.rb test/factbase/indexed/test_indexed_query.rb` — passed.

### DIFF
--- a/lib/factbase/indexed/indexed_gt.rb
+++ b/lib/factbase/indexed/indexed_gt.rb
@@ -16,6 +16,7 @@ class Factbase::IndexedGt
     prop = op1.to_s
     target = op2.is_a?(Symbol) ? params[op2.to_s]&.first : op2
     return maps || [] if target.nil?
+    return unless sortable?(maps, prop)
     key = [maps.object_id, prop, :facts]
     @idx[key] ||= { facts: [], count: 0 }
     entry = @idx[key]
@@ -39,6 +40,13 @@ class Factbase::IndexedGt
     end
     entry[:facts].sort_by! { |pair| pair[0] }
     entry[:count] = facts.size
+  end
+
+  def sortable?(maps, prop)
+    maps.to_a.flat_map { |fact| fact[prop] || [] }.sort
+    true
+  rescue ArgumentError
+    false
   end
 
   def _search(entry, target)
--- a/lib/factbase/indexed/indexed_gte.rb
+++ b/lib/factbase/indexed/indexed_gte.rb
@@ -16,6 +16,7 @@ class Factbase::IndexedGte
     prop = op1.to_s
     target = op2.is_a?(Symbol) ? params[op2.to_s]&.first : op2
     return maps || [] if target.nil?
+    return unless sortable?(maps, prop)
     key = [maps.object_id, prop, :facts]
     @idx[key] ||= { facts: [], count: 0 }
     entry = @idx[key]
@@ -39,6 +40,13 @@ class Factbase::IndexedGte
     end
     entry[:facts].sort_by! { |pair| pair[0] }
     entry[:count] = facts.size
+  end
+
+  def sortable?(maps, prop)
+    maps.to_a.flat_map { |fact| fact[prop] || [] }.sort
+    true
+  rescue ArgumentError
+    false
   end
 
   def _search(entry, target)
--- a/lib/factbase/indexed/indexed_lt.rb
+++ b/lib/factbase/indexed/indexed_lt.rb
@@ -16,6 +16,7 @@ class Factbase::IndexedLt
     prop = op1.to_s
     target = op2.is_a?(Symbol) ? params[op2.to_s]&.first : op2
     return maps || [] if target.nil?
+    return unless sortable?(maps, prop)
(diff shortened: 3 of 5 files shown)


yegor256:
Merged. Thanks — falling back to the ordinary comparison path rather than caching a failed sort is the right shape, and asserting identical messages across plain and indexed is what makes it a real regression.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+728](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #822, opened 2026-09-07
#806: wrap every fact the cached query yields

### LINKED ISSUE #806: `CachedFactbase` keeps answering from a stale cache after a write through a parameterized query
## What happens

A write made through a query that carries a parameter does not invalidate the cache, so every
query cached before it keeps answering from the old snapshot for the life of the factbase.

```ruby
fb = Factbase::CachedFactbase.new(Factbase.new)
fb.insert.then { |f| f.foo = 1 }
puts fb.query('(gt foo 5)').each.to_a.size           # 0, correct, and now cached
fb.query('(eq foo $x)').each(fb, x: [1]) { |f| f.foo = 9 }
puts fb.query('(always)').each.to_a.map(&:to_s)
puts fb.query('(gt foo 5)').each.to_a.size
```

```
before=0
["[ foo: [1, 9] ] + {}"]
after=0
```

The fact really did change, and the answer is still the old one. The same run against a plain
factbase and against `IndexedFactbase` gives `before=0 after=1`, which is what it should be.

## Why

https://github.com/yegor256/factbase/blob/master/lib/factbase/cached/cached_factbase.rb

```ruby
def query(term, maps = nil)
  term = to_term(term) if term.is_a?(String)
  q = @origin.query(term, maps)
  q = Factbase::CachedQuery.new(q, @cache, self) unless term.abstract?
  q
end
```

A term holding a `$` is `abstract?`, so the wrapper is skipped and the raw query comes back. That
query yields plain facts rather than `Factbase::CachedFact`, and `CachedFact` is the only thing
that clears the cache when a property is written. So the write happens and nothing tells the cache.

`IndexedFactbase` does not have the hole: `IndexedQuery#each` wraps every yielded fact in
`IndexedFact` whether the term is abstract or not.

This is not #717, which was about the cache being shared with a transaction.

## What should happen

A write through any query a `CachedFactbase` returns should invalidate the cache. Wrapping the
yielded facts always, and skipping only the caching of the result for an abstract term, would do
it, which is the shape the indexed decorator already uses.

morphqdd:
A term holding a `$` is abstract, so the query came back unwrapped, and the facts it yielded were plain ones rather than `Factbase::CachedFact`. Only `CachedFact` clears the cache when a property is written, so a write through such a query left every earlier answer in place.

Every query is wrapped now, and the term decides only whether the result may be cached, which is the shape the indexed decorator already uses.

Closes #806

### DIFF
--- a/lib/factbase/cached/cached_factbase.rb
+++ b/lib/factbase/cached/cached_factbase.rb
@@ -49,9 +49,7 @@ class Factbase::CachedFactbase
   # @param [Array<Hash>] maps Possible maps to use
   def query(term, maps = nil)
     term = to_term(term) if term.is_a?(String)
-    q = @origin.query(term, maps)
-    q = Factbase::CachedQuery.new(q, @cache, self) unless term.abstract?
-    q
+    Factbase::CachedQuery.new(@origin.query(term, maps), @cache, self, cacheable: !term.abstract?)
   end
 
   # Run an ACID transaction.
--- a/lib/factbase/cached/cached_query.rb
+++ b/lib/factbase/cached/cached_query.rb
@@ -17,10 +17,13 @@ class Factbase::CachedQuery
   # Constructor.
   # @param [Factbase::Query] origin Original query
   # @param [Hash] cache The cache
-  def initialize(origin, cache, fb)
+  # @param [Factbase] fb The factbase
+  # @param [Boolean] cacheable Whether the result of the query may be cached
+  def initialize(origin, cache, fb, cacheable: true)
     @origin = origin
     @cache = cache
     @fb = fb
+    @cacheable = cacheable
   end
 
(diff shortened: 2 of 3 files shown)


morphqdd:
@yegor256, CI is green here. Please take a look.

yegor256:
Merged, thanks! Wrapping always and letting the term decide only cacheability is cleaner than the two-shaped return it replaces.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+7200](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #824, opened 2026-09-08
#823: keep the sprintf context when an operand is nil

### LINKED ISSUE #823: sprintf loses its contextual message for a missing property, because the rescue misses TypeError
## What happens

`Factbase::Sprintf#formatted` wraps `Kernel#format` and catches one exception class:

```ruby
def formatted(fmt, ops)
  format(*([fmt] + ops))
rescue ArgumentError => e
  raise(RuntimeError, "Cannot format #{ops.inspect} with '#{fmt}' in (sprintf ...): #{e.message}")
end
```

`format` raises `ArgumentError` for a value it cannot convert and for a broken format string, and `TypeError` for `nil`. The `nil` case is the one that slips past, and it is also the most likely one, because an operand is `nil` exactly when the fact does not carry that property.

## Reproduction

```ruby
fb = Factbase.new
f = fb.insert
f.n = 42
f.s = 'hi'
```

| query | message |
| --- | --- |
| `(as r (sprintf "%d" s))` | `Cannot format ["hi"] with '%d' in (sprintf ...): invalid value for Integer(): "hi"` |
| `(as r (sprintf "%s %s" n))` | `Cannot format [42] with '%s %s' in (sprintf ...): too few arguments` |
| `(as r (sprintf "%" n))` | `Cannot format [42] with '%' in (sprintf ...): incomplete format specifier` |
| `(as r (sprintf "%d" missing))` | **`can't convert nil into Integer`** |
| `(as r (sprintf "%f" missing))` | **`can't convert nil into Float`** |

The first three name the values and the format. The last two name neither, because they never reach the `rescue`: the `TypeError` travels up and is wrapped further out, which appends `at (sprintf '%d' missing)` and nothing else.

So the failure that says the least is the one a reader is most likely to hit, and the one where knowing the format and the operands would help most: the message does not say which of several operands was missing.

## Suggested fix

One word:

```ruby
rescue ArgumentError, TypeError => e
  raise(RuntimeError, "Cannot format #{ops.inspect} with '#{fmt}' in (sprintf ...): #{e.message}")
```

With that, `(sprintf "%d" missing)` says `Cannot format [nil] with '%d' in (sprintf ...): can't convert nil into Integer`, which points straight at the absent property.

## Related

#579 introduced this `rescue` for exactly this reason, and covered `ArgumentError`. `TypeError` is the other half of what `format` throws, and it is the half a missing property produces.

gemshrine:
Fixes #823

`formatted` rescued only `ArgumentError`, but `Kernel#format` answers `nil` with a `TypeError`, and an operand is `nil` exactly when the fact does not carry that property, which is the common case. The `TypeError` went past the rescue and reached the caller as bare `can't convert nil into Integer`, naming neither the format nor the operands, while every other failure in the same method arrives as `Cannot format [...] with '...' in (sprintf ...)`.

Rescuing `TypeError` alongside `ArgumentError` puts the missing-property case in the same shape as the rest. Nothing else changes: the class raised is still `RuntimeError`, and the three existing cases keep their messages.

The new test covers `%d` and `%f` with an absent property. It fails on master, where the raised class is `TypeError`.

### DIFF
--- a/lib/factbase/terms/sprintf.rb
+++ b/lib/factbase/terms/sprintf.rb
@@ -33,7 +33,7 @@ class Factbase::Sprintf < Factbase::TermBase
 
   def formatted(fmt, ops)
     format(*([fmt] + ops))
-  rescue ArgumentError => e
+  rescue ArgumentError, TypeError => e
     raise(RuntimeError, "Cannot format #{ops.inspect} with '#{fmt}' in (sprintf ...): #{e.message}")
   end
 end
--- a/test/factbase/terms/test_sprintf.rb
+++ b/test/factbase/terms/test_sprintf.rb
@@ -31,6 +31,18 @@ class TestSprintf < Factbase::Test
     assert_includes(e.message, 'invalid value for Integer')
   end
 
+  def test_names_the_format_when_an_operand_is_absent
+    [['%d', :absent], ['%f', :absent]].each do |ops|
+      t = Factbase::Sprintf.new(ops)
+      e =
+        assert_raises(RuntimeError) do
+          t.evaluate(fact, [], Factbase.new)
+        end
+      assert_includes(e.message, "Cannot format [nil] with '#{ops[0]}' in (sprintf ...):")
+      assert_includes(e.message, 'nil')
+    end
+  end
+
   def test_rejects_missing_format_operand
     t = Factbase::Sprintf.new(['%s %s', 'hello'])
     e =


yegor256:
Merged, thanks! Rescuing `TypeError` next to `ArgumentError` puts the missing-property case in the same shape as the rest.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1844](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #853, opened 2026-09-09
#829: let `matches` test every value of the property

### LINKED ISSUE #829: `matches` refuses a multi-valued property, while `contains`, `starts_with` and `ends_with` accept one
`lib/factbase/terms/matches.rb:29`:

```ruby
raise(RuntimeError, 'Exactly one string is expected') unless str.size == 1
```

README says `(matches v s)` is true if any `v` matches the `s` regular expression. `contains`, `starts_with` and `ends_with` all go through `Factbase::Compare`, which does `lefts.any? { |l| rights.any? { ... } }` and works on a property with two values. `matches` is the only one that throws.

What happens, on a fact with `foo = 'aa'` and `foo = 'bb'`:

```
contains: 1
starts_with: 1
ends_with: 1
matches: "Exactly one string is expected" at (matches foo 'bb')
```

The message names neither the term nor the property.

What should happen: `matches` tests every value of the left operand and is true if any of them matches, like its three siblings.

morphqdd:
`matches` refused a property with more than one value, while `contains`,
`starts_with` and `ends_with` all test every value and pass if any of them
matches. `matches` now behaves the same.

Closes #829

### DIFF
--- a/lib/factbase/terms/matches.rb
+++ b/lib/factbase/terms/matches.rb
@@ -21,16 +21,16 @@ class Factbase::Matches < Factbase::TermBase
   # @param [Factbase::Fact] fact The fact
   # @param [Array<Factbase::Fact>] maps All maps available
   # @param [Factbase] fb Factbase to use for sub-queries
-  # @return [Boolean] True if the string matches the regexp, false otherwise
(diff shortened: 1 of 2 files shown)


morphqdd:

PR #857, opened 2026-09-09
#835: drop the cache when facts are imported

### LINKED ISSUE #835: `CachedFactbase#import` leaves the cache in place, so imported facts stay invisible
`insert` marks the cache dirty in `lib/factbase/cached/cached_factbase.rb:34`, `delete!` clears it in `cached_query.rb:65`, and a property write clears it in `cached_fact.rb:28`. `import` is not overridden at all, so it is delegated straight to the origin and the cache survives it.

What happens:

```ruby
src = Factbase.new
src.insert.foo = 42
fb2 = Factbase::CachedFactbase.new(Factbase.new)
fb2.query('(exists foo)').each.to_a   # caches the empty answer
fb2.import(src.export)
puts fb2.query('(exists foo)').each.to_a.size
```

```
no prior query: 1 (expected 1)
prior query   : 0 (expected 1), fb.size=1
```

Every query that ran before the import keeps answering from the stale entry.

#467 was about changes made behind the decorator's back. This write goes through the decorator's own public API.

What should happen: `import` invalidates the cache, the way `insert` and `delete!` do.

morphqdd:
`import` was delegated straight to the origin, so every query answered before
the import kept returning its stale entry and the imported facts stayed
invisible. It clears the cache now, the way `insert` and `delete!` do.

Closes #835

### DIFF
--- a/lib/factbase/cached/cached_factbase.rb
+++ b/lib/factbase/cached/cached_factbase.rb
@@ -35,6 +35,13 @@ class Factbase::CachedFactbase
     Factbase::CachedFact.new(@origin.insert, @cache, fresh: true)
   end
 
+  # Import facts from a binary buffer.
+  # @param [String] data The data to import
+  def import(data)
+    @cache.clear
+    @origin.import(data)
+  end
+
   # Convert a query to a term.
   # @param [String] query The query to convert
   # @return [Factbase::Term] The term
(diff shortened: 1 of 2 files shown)


morphqdd:

yegor256:
Merged, thanks! Clearing the cache before delegating is the right order, and the test proves the query is not the one that primed it.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+8092](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #863, opened 2026-09-09
#830: explain a `defn` body that is not valid Ruby

### LINKED ISSUE #830: A malformed `defn` body raises `SyntaxError`, which no `rescue StandardError` catches
`defn` builds Ruby source from its second operand and calls `Kernel#eval` in `lib/factbase/terms/defn.rb:34-37`. If the operand is not valid Ruby, `eval` raises `SyntaxError`, which is a `ScriptError`, not a `StandardError`.

What happens: neither of the two rescue clauses in `Factbase::Term#evaluate` catches it, and the caller's own `rescue => e` does not catch it either.

```ruby
begin
  fb.query('(defn brk "1 +")').each.to_a
rescue StandardError => e
  puts "caught: #{e.message}"
end
```

```
ESCAPED SyntaxError: .../lib/factbase/terms/defn.rb:39: syntax errors found
```

Only an explicit `rescue Exception` catches it. The message quotes the generated `def brk(fact, maps, fb)` source and points at `defn.rb`, not at the query that caused it. Every other bad operand in the gem produces a contextual `RuntimeError`.

What should happen: `defn` rescues `SyntaxError` and re-raises it as a normal error naming the term and the offending body, for example `Cannot define the term 'brk', its body is not valid Ruby: ...`.

morphqdd:
A malformed `defn` body raised `SyntaxError`, which is a `ScriptError`, so no
`rescue StandardError` in the gem or in the caller could catch it. It is turned
into an `ArgumentError` naming the term now.

Closes #830

### DIFF
--- a/lib/factbase/terms/defn.rb
+++ b/lib/factbase/terms/defn.rb
@@ -30,12 +30,16 @@ class Factbase::Defn < Factbase::TermBase
     raise(ArgumentError, "Term '#{fn}' is already defined") if Factbase::Term.private_method_defined?(fn, false)
     raise(ArgumentError, "Term '#{fn}' is already defined") if Factbase::Term::TERMS.key?(fn)
     raise(ArgumentError, "The '#{fn}' is a bad name for a term") unless fn.match?(/^[a-z_]+$/)
-    # rubocop:disable Security/Eval
-    eval(
-      "class Factbase::Term\nprivate\ndef #{fn}(fact, maps, fb)\n#{@operands[1]}\nend\nend",
-      binding, __FILE__, __LINE__ - 1
-    )
-    # rubocop:enable Security/Eval
+    begin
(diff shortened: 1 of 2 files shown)


morphqdd:

yegor256:
Merged, thanks! Turning the `SyntaxError` into an `ArgumentError` is what makes it catchable by the callers that rescue `StandardError`.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+8136](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #864, opened 2026-09-09
#826: remember which quote opened a string literal

### LINKED ISSUE #826: A string literal ends at a quote of the other kind, so a quote inside it breaks or silently splits the query
In `lib/factbase/syntax.rb:101-113` the tokenizer keeps `quotes = ['\'', '"']` and a single boolean `string` flag. It never records which quote opened the literal, so a quote of the other kind inside the literal toggles the flag.

What happens, with an odd number of intruding quotes the query dies:

```ruby
fb.query(%q{(eq foo "Jeff's")}).each.to_a
fb.query(%q{(eq foo 'a"b')}).each.to_a
```

```
(eq foo "Jeff's")  -> ERR String not closed
(eq foo 'a"b')     -> ERR String not closed
```

With an even number the query does not fail, it quietly means something else. The characters between the two intruding quotes are processed as if they were outside the string, so a space there flushes the accumulator and the literal is cut into two tokens:

```ruby
Factbase::Syntax.new(%q{(eq foo 'a" "b')}).send(:to_tokens)
```

```
[:open, :eq, :foo, "a", "b", :close]
Too many (3) operands for '==' (2 expected)
```

What should happen: a literal ends only at the same quote character that opened it. `(eq foo "Jeff's")` should match the value `Jeff's`, and `(eq foo 'a" "b')` should be one string operand.

morphqdd:
The tokenizer kept one boolean for "inside a string", so a quote of the other
kind closed the literal: `(eq foo "Jeff's")` died with "String not closed" and
an even number of intruding quotes silently cut the literal into two tokens.
It remembers the opening quote now and only that one closes the literal.

Overlaps with open PRs #843 and #813 on `lib/factbase/syntax.rb`, so this goes up as a draft.

Closes #826

### DIFF
--- a/lib/factbase/syntax.rb
+++ b/lib/factbase/syntax.rb
@@ -100,20 +100,22 @@ class Factbase::Syntax
     acc = ''
     quotes = ['\'', '"']
     spaces = [' ', ')', "\n", "\t", "\r"]
-    string = false
+    opener = nil
     comment = false
     @query.to_s.chars.each do |c|
-      comment = true if !string && c == '#'
+      comment = true if opener.nil? && c == '#'
       comment = false if comment && c == "\n"
       next if comment
       if quotes.include?(c)
-        if string && acc[-1] == '\\'
+        if !opener.nil? && acc[-1] == '\\'
           acc = acc[0..-2]
-        else
-          string = !string
+        elsif opener.nil?
+          opener = c
+        elsif opener == c
+          opener = nil
         end
       end
-      if string
+      if opener
         acc += c
         next
       end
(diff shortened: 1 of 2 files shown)


yegor256:
Merged, thanks! Remembering the opening quote instead of a boolean is the correct shape, and the two tests pin both directions.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+8246](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #865, opened 2026-09-09
#825: let an opening bracket terminate a token

### LINKED ISSUE #825: The tokenizer does not treat `(` as a separator, so `(and(eq foo 1))` becomes one unknown term
The list of characters that terminate an accumulated token in `lib/factbase/syntax.rb:102` is `[' ', ')', "\n", "\t", "\r"]`. The opening bracket is not there.

What happens:

```ruby
Factbase::Syntax.new('(and(eq foo 1))').send(:to_tokens)
```

```
[:open, :open, :andeq, :foo, 1, :close, :close]
```

`and` is never flushed, so it is glued to `eq` into one identifier that does not exist, and it is emitted after `:open` is already pushed. Parsing then fails with:

```
Too many terms (6 != 7) (ArgumentError)
```

The message points nowhere near the missing space.

This is the same defect that #601 fixed for `\t`, `\n` and `\r`. The `(` was left out of the same list.

What should happen: `(` must terminate the current token exactly like `)` does, so that `(and(eq foo 1))` parses the same way as `(and (eq foo 1))`.

morphqdd:
`(` was missing from the list of characters that end a token, so `(and(eq foo 1))`
glued `and` and `eq` into one unknown identifier and failed with a message that
pointed nowhere near the missing space. It ends a token now, the way `)` does.

Overlaps with open PRs #843 and #813 and with #864 on `lib/factbase/syntax.rb`, so this goes up as a draft.

Closes #825


yegor256:
Merged, thanks! The tokenizer now ends a token at an opening bracket, so `(and(eq foo 1))` parses the way it reads.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+11334](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #866, opened 2026-09-12
#844: index a repeated value once

### LINKED ISSUE #844: IndexedFactbase returns the same fact once per duplicate property value
`Factbase::IndexedFactbase` can return one result more than once when a fact stores the same value repeatedly. The equality index appends the fact once for every value in `lib/factbase/indexed/indexed_eq.rb:38-44`, but only calls `uniq!` when the query parameter resolves to multiple keys (`:27`). A scalar equality query therefore preserves duplicate references.

Reproduction:

```ruby
maps = [{ 'a' => [2, 2] }, { 'a' => [1] }]
plain = Factbase.new(maps.map(&:dup))
indexed = Factbase::IndexedFactbase.new(Factbase.new(maps.map(&:dup)))
plain.query('(eq a 2)').each.to_a.size   # 1
indexed.query('(eq a 2)').each.to_a.size # 2
```

The indexed and plain implementations should have identical set semantics. Deduplicate the scalar-key matches (or otherwise ensure each fact is yielded once) and add a regression test for repeated values.

gemshrine:
Fixes #844.

A fact that stores the same value twice was appended to that value's bucket twice, so `(eq a 2)` over `{ 'a' => [2, 2] }` returned it twice through the index while a plain `Factbase` returned it once. The `uniq!` in `predict` only ran when the parameter resolved to several keys, so the scalar case kept the duplicate.

The fix is in `_feed`: a fact enters each bucket once, no matter how often the value repeats. That keeps the deduplication where it costs nothing, instead of on every read of the bucket, and the multi-key `uniq!` stays as it is, since there a fact can legitimately sit in two different buckets.

### DIFF
--- a/lib/factbase/indexed/indexed_eq.rb
+++ b/lib/factbase/indexed/indexed_eq.rb
@@ -38,7 +38,7 @@ class Factbase::IndexedEq
   def _feed(facts, entry, operand)
     return unless entry[:count] < facts.size
     facts[entry[:count]..].each do |m|
-      m[operand]&.each do |v|
+      m[operand]&.uniq&.each do |v|
         entry[:facts][v] ||= []
         entry[:facts][v] << m
       end
--- a/test/factbase/indexed/test_indexed_eq.rb
+++ b/test/factbase/indexed/test_indexed_eq.rb
@@ -2,6 +2,7 @@
 
 require_relative '../../../lib/factbase'
 require_relative '../../../lib/factbase/indexed/indexed_eq'
+require_relative '../../../lib/factbase/indexed/indexed_factbase'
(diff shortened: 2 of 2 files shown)


yegor256:
Merged, thanks! Comparing the indexed and the plain factbase in the test is the right oracle for this.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2874](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #867, opened 2026-09-12
#845: keep the insertion order in gte and lte

### LINKED ISSUE #845: Indexed range queries reorder facts compared with Factbase
The indexed implementations of `(gte ...)` and `(lte ...)` return matching facts in value-sorted index order instead of the insertion order returned by a regular `Factbase`. `lib/factbase/indexed/indexed_gte.rb:54-59` and `indexed_lte.rb:54-59` map the sorted index directly to facts; the sibling `gt`/`lt` implementations explicitly restore the original position, so range behavior is inconsistent.

Reproduction:

```ruby
maps = [{ 'a' => [3] }, { 'a' => [0] }, { 'a' => [2] }]
plain = Factbase.new(maps.map(&:dup))
indexed = Factbase::IndexedFactbase.new(Factbase.new(maps.map(&:dup)))
plain.query('(gte a 1)').each.to_a.map { |f| f['a'].first }   # [3, 2]
indexed.query('(gte a 1)').each.to_a.map { |f| f['a'].first } # [2, 3]

maps = [{ 'a' => [2] }, { 'a' => [0] }, { 'a' => [3] }]
# plain `(lte a 2)` is [2, 0], while indexed is [0, 2]
```

Query decorators should not change result ordering merely because an index is enabled. Restore source positions for `gte` and `lte` (as `IndexedGt` and `IndexedLt` already do) and add regression tests for all four range operators.

gemshrine:
Fixes #845.

`IndexedGte` and `IndexedLte` handed back the facts in the order of the sorted index, so turning the index on changed the answer of a query: `(gte a 1)` over `[3, 0, 2]` gave `[2, 3]` instead of the `[3, 2]` a plain `Factbase` returns.

`IndexedGt` and `IndexedLt` already keep the source position in the index entry and sort by it before returning. Both siblings now do the same, so all four range operators agree with a plain `Factbase`.

There were no test files for the two of them; they exist now, each with the order test its `gt`/`lt` sibling has and a direct comparison against a plain `Factbase` over the same maps. All four assertions fail on master.

### DIFF
--- a/lib/factbase/indexed/indexed_gte.rb
+++ b/lib/factbase/indexed/indexed_gte.rb
@@ -33,9 +33,9 @@ class Factbase::IndexedGte
 
   def _feed(facts, entry, prop)
     return unless entry[:count] < facts.size
-    facts[entry[:count]..].each do |fact|
+    facts[entry[:count]..].each_with_index do |fact, i|
       fact[prop]&.each do |v|
-        entry[:facts] << [v, fact]
+        entry[:facts] << [v, fact, entry[:count] + i]
       end
     end
     entry[:facts].sort_by! { |pair| pair[0] }
@@ -52,7 +52,7 @@ class Factbase::IndexedGte
   def _search(entry, target)
     idx = entry[:facts].bsearch_index { |v, _| v >= target }
     return [] if idx.nil?
-    facts = entry[:facts][idx..].map { |_, f| f }
+    facts = entry[:facts][idx..].sort_by { |pair| pair[2] }.map { |pair| pair[1] }
     facts.uniq!(&:object_id)
     facts
   end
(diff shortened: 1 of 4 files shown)


yegor256:
Merged, thanks! Comparing against a plain factbase is the right oracle for this. One nit for a follow-up: the SPDX headers in the two new test files sit below the `require_relative` lines instead of at the top.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +7 for the 98 hits-of-code that you wrote; -7 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3550](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #868, opened 2026-09-12
#847: cache the negation per set of params

### LINKED ISSUE #847: Indexed negation caches results across different query parameters
## The bug
`Factbase::IndexedNot` caches its predicted result using only the maps object, sub-term, and operator (`lib/factbase/indexed/indexed_not.rb:16-21`). The cache key omits query parameters. A negated term that references `$` therefore reuses the first parameter's matching set on later calls.

## Reproduction
```ruby
origin = Factbase.new
origin.insert.foo = 1
origin.insert.foo = 2
fb = Factbase::IndexedFactbase.new(origin)
query = fb.query('(not (eq foo $x))')
query.each(fb, {'x' => 1}).to_a.map { |f| f['foo'].first } # => [2]
query.each(fb, {'x' => 2}).to_a.map { |f| f['foo'].first } # => [] (expected [1])
```

## Expected
The prediction cache must include the effective parameters (or be bypassed for parameterized terms), so each invocation returns the same facts as the plain Factbase query.

gemshrine:
Fixes #847.

`IndexedNot` keyed its prediction on the maps, the sub-term and the operator, but not on the parameters, so `(not (eq foo $x))` answered every later call with the set it computed for the first `$x`. In the reproduction from the issue the second call returned nothing instead of `[1]`.

The key now carries a snapshot of the parameters. At query time they arrive as a `Factbase::Tee` rather than a `Hash` (its `to_a` is delegated to the wrapped fact, which is empty, so the naive `params.to_a` snapshot is the same for every call), so `snapshot` reads them through `all_properties` and `$`-lookup, and still handles a plain `Hash`, which is what the unit tests pass.

The test runs the reproduction, then repeats the first parameter to show the cache is still used and still correct. It fails on master.

### DIFF
--- a/lib/factbase/indexed/indexed_not.rb
+++ b/lib/factbase/indexed/indexed_not.rb
@@ -12,7 +12,7 @@ class Factbase::IndexedNot
 
   def predict(maps, fb, params)
     sub = @term.operands.first
-    key = [maps.object_id, sub, @term.op]
+    key = [maps.object_id, sub, @term.op, snapshot(params)]
     @idx[key] ||= { facts: nil, count: 0, yes_set: nil }
     entry = @idx[key]
     _feed(maps.to_a, entry) do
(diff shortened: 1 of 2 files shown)


yegor256:
Merged, thanks! Keying the cache on the parameters is the right fix, and the test pins it down well.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3526](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #869, opened 2026-09-12
#846: write a new fact while holding the monitor

### LINKED ISSUE #846: SyncFactbase releases the monitor before a new fact is populated
`Factbase::SyncFactbase` is documented as a thread-safe factbase, but `insert` only synchronizes the call that allocates the empty fact (`lib/factbase/sync/sync_factbase.rb:34-38`). The returned `Fact` is the raw object from `@origin`, so the caller populates its properties after the monitor has already been released. A concurrent query can observe the empty or partially populated fact, and two writers can mutate it without synchronization.

```ruby
sync = Factbase::SyncFactbase.new(Factbase.new)
fact = sync.insert                 # monitor is released here
fact.name = 'value'                # unsynchronized mutation
```

The decorator should return a fact wrapper that performs property writes while holding the same monitor (or expose insertion/population as one synchronized operation), and a concurrent read/write regression test should demonstrate that no partial fact is observable.

gemshrine:
Fixes #846.

`SyncFactbase#insert` held the monitor only while the empty fact was allocated and then handed back the raw fact of the origin. Everything the caller wrote afterwards, which is where the fact actually gets its content, happened with no synchronization at all: two writers could mutate the same fact at once, and a reader could walk it mid-write.

`insert` now returns a `Factbase::SyncFact`, a thin wrapper that takes the same monitor around every call it forwards, so the writes that follow an insert are as synchronized as the insert itself. That is the first of the two options in the issue; the second one (insertion and population as a single operation) would change the signature of `insert` for every caller.

Note that this makes each write atomic, not the whole population of a fact: a reader can still see a fact that has `a` but not yet `b`. Making the sequence atomic needs a block form of `insert`, which is a larger change, and I did not want to smuggle it in here.

The test in `test_sync_factbase.rb` holds the monitor in the main thread and shows that a write from another thread does not go through until the monitor is released; on master it goes through immediately. `test_sync_fact.rb` covers the wrapper itself.

### DIFF
--- a/lib/factbase/sync/sync_fact.rb
+++ b/lib/factbase/sync/sync_fact.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Yegor Bugayenko
+# SPDX-License-Identifier: MIT
+
+require 'others'
+require_relative '../../factbase'
+
+# A single fact, which is only touched while holding the monitor.
+#
+# Author:: Yegor Bugayenko (yegor256@gmail.com)
+# Copyright:: Copyright (c) 2024-2026 Yegor Bugayenko
+# License:: MIT
+class Factbase::SyncFact
+  # Ctor.
+  # @param [Factbase::Fact] origin The original fact
+  # @param [Monitor] monitor The monitor to hold while reading or writing
+  def initialize(origin, monitor)
+    @origin = origin
+    @monitor = monitor
+  end
+
+  def to_s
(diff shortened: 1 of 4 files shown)


yegor256:
CI is green, but I'd rather not merge this one unreviewed. `insert` now answers with a `Factbase::SyncFact` rather than the `Factbase::Fact` it used to, and every method goes through `others` delegation, so any caller that checks the class of what `insert` returned, or relies on the fact's own `method_missing`, changes behaviour. Also, `require_relative` sits inside `insert` rather than at the top of the file. Assigning to @yegor256 for a decision.

---

PR #908, opened 2026-09-15
#889: look for the index before asking how many there are

### LINKED ISSUE #889: `at` with an absent property says that the term `at` is not defined
## What happens

`at` asks for the size of its index before it checks whether the index is there at all
(`lib/factbase/term.rb:259-263`):

```ruby
i = _values(0, fact, maps, fb)
raise(RuntimeError, "Too many values (#{i.size}) at first position, one expected") unless i.size == 1
i = i[0]
return if i.nil?
```

`_values` answers `nil` for an absent property, so `i.size` throws, and the catch-all at
`term.rb:210` turns it into a message about a term that does not exist:

```ruby
fb.query('(eq foo (at absent_prop multi))').each.to_a.size
# RuntimeError: "Probably the term 'at' is not defined at (at absent_prop multi):
#  undefined method 'size' for nil"
```

`at` is defined. The reader is sent looking for a missing term while the real cause is an absent
property. `Arithmetic#evaluate` does the same two steps in the right order
(`terms/arithmetic.rb:25`, `return if lefts.nil?` before `lefts.size`), and so does `either`.

## What should happen

The nil check should come before the size check, and `at` should answer nothing for an absent
index, the way arithmetic does.

morphqdd:
`at` asked for the size of its index before checking whether the index was there, so an absent property made `i.size` throw and the catch-all reported it as a term that does not exist. `at` is defined, and the reader was sent looking for the wrong thing.

The nil check now comes first, the way `Arithmetic#evaluate` and `either` already do it.

Closes #889


yegor256:
Merged, thanks! A one-line guard with a test that names the behaviour is exactly the right size for this.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+10456](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #909, opened 2026-09-15
#890: let a wrong operand on either side give the same message

### LINKED ISSUE #890: `contains`, `starts_with` and `ends_with` lose their own message when the right operand is not a String
## What happens

`Factbase::Compare` builds a careful message when the two operands cannot be compared
(`lib/factbase/terms/compare.rb:48-51`), but the rescue that leads to it lists only two error
classes (`compare.rb:47`):

```ruby
rescue ArgumentError, NoMethodError => e
```

`String#include?(7)` raises `TypeError`, which is not in the list, so the message never fires:

```
(contains name 7)      => RuntimeError: no implicit conversion of Integer into String
(starts_with name 7)   => RuntimeError: no implicit conversion of Integer into String
(ends_with name 7)     => RuntimeError: no implicit conversion of Integer into String
(contains foo '4')     => RuntimeError: Cannot compare 42 (Integer) with "4" (String)
                          using (compare include?): undefined method 'include?' ...
```

The left operand case names the values, their classes and the operator. The right operand case
names none of them, although it is the same mistake seen from the other side.

This is the same defect that was accepted as #823 for `sprintf`, in a different file.

## What should happen

`TypeError` should be in the rescue list at `compare.rb:47`, so that a wrong operand on either
side produces the message the class already writes.

morphqdd:
`Factbase::Compare` writes a message naming both values, their classes and the operator, but the rescue that leads to it listed only `ArgumentError` and `NoMethodError`. `String#include?(7)` raises `TypeError`, so `contains`, `starts_with` and `ends_with` lost that message whenever the wrong operand was on the right.

`TypeError` is now in the list, so the same mistake reads the same from either side.

Closes #890


yegor256:
Merged, thanks! Adding `TypeError` to the rescue keeps all three predicates speaking with one voice.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+10476](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #910, opened 2026-09-15
#891: say that env wants a string for its name

### LINKED ISSUE #891: `env` with a non-String name says that the term `env` is not defined
## What happens

`env` upcases its first operand without looking at what it is
(`lib/factbase/terms/env.rb:27`):

```ruby
value = ENV.fetch(n[0].upcase, nil)
```

So a non-String name throws `NoMethodError`, which the catch-all at `term.rb:210` reports as a
term that does not exist:

```ruby
g = fb.insert; g.n = 'fb_test_var'; g.i = 5
fb.query('(eq n (env i "d"))').each.to_a.size
# RuntimeError: "Probably the term 'env' is not defined at (env i 'd'):
#  undefined method 'upcase' for an instance of Integer"
```

`env` is defined. #655 covers the operand resolving to nothing; the non-String case is a
separate branch of the same line and is still unguarded.

## What should happen

`env` should say that a string is expected as its first argument, the way `as`
(`terms/as.rb:23`), `join` (`join.rb:24`) and `assert` (`assert.rb:27`) already do for theirs.

morphqdd:
`env` upcased its first operand without looking at what it was, so a non-string name threw `NoMethodError` and the catch-all reported it as a term that does not exist. `env` is defined, and the reader was pointed at the wrong thing.

It now says a string is expected, the way `as`, `join` and `assert` already do for their own arguments.

Closes #891


yegor256:
Merged, thanks! Failing at the argument instead of deep inside `ENV.fetch` is the better error.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+10496](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #914, opened 2026-09-15
#880: keep no fact whose block has thrown

### LINKED ISSUE #880: `Factbase::Pre` keeps the fact whose block has thrown
## What happens

`Factbase::Pre#insert` puts the fact into the factbase first and runs the block afterwards
(`lib/factbase/pre.rb:34-38`):

```ruby
def insert
  f = @fb.insert
  @block.call(f, self)
  f
end
```

When the block throws, the fact stays:

```ruby
p = Factbase::Pre.new(Factbase.new) { |f, _| raise('nope') }
p.insert rescue nil
p.size   # => 1
```

What is left behind is an empty fact without any of the properties the decorator exists to put
there. Its own rdoc (`pre.rb:12-13`) says the decorator is for putting some properties into
every fact that gets into the factbase.

## What should happen

A fact whose block has thrown should not remain in the factbase: either take it out when the
block fails, or run the block before the fact becomes visible.

morphqdd:
`Pre#insert` inserted the fact and then ran the block on it. A block that threw left the half built fact in the factbase, visible to every later query, with whatever properties it had managed to set.

The insert and the block now happen inside one transaction, so a failure takes the fact with it. Inside an enclosing transaction, where a nested one is not allowed, the outer transaction already gives the same rollback, and the new flag keeps the two paths apart.

All 647 tests stay green.

Closes #880


yegor256:
CI is green and the tests read well, but this one needs your judgement rather than mine.

`Factbase::Pre#insert` now opens a transaction of its own whenever it is not already inside one, and adds a public `inside:` keyword to the constructor to track that. Opening an implicit transaction inside a helper is the exact shape that zerocracy/fbe#843 reports as a problem for `Fbe.overwrite`: a caller who is already inside a transaction of their own ends up nesting. Whether that is safe here depends on how `Factbase#txn` composes, which is a decision about the core rather than about `Pre`.

Assigning to you.

---

PR #917, opened 2026-09-15
#892: cut a long value down to the limit, not past it

### LINKED ISSUE #892: `Factbase::Logged` shortens a value into something longer than `MAX_LENGTH`
## What happens

`Factbase::Logged::Fact` shortens a long value like this
(`lib/factbase/logged.rb:98,126`):

```ruby
MAX_LENGTH = 64
s = "#{s[0..(MAX_LENGTH / 2)]}...#{s[(-MAX_LENGTH / 2)..]}" if s.length > MAX_LENGTH
```

`s[0..32]` is 33 characters, `...` is three more, and `s[-32..]` is 32, which makes 68. The
result is longer than the limit it is cut to. A string is rendered with `inspect` first
(line 125), so a 63-character value is already 65 with its quotes, crosses the threshold, and
comes out of the shortening at 68 characters:

```
value length     = 63
inspect length   = 65   (MAX_LENGTH = 64)
rendered length  = 68
```

Line 124 is also dead for strings: `s = v.to_s` is overwritten unconditionally by line 125.

## What should happen

The two halves should be `(MAX_LENGTH - 3) / 2` each, so that nothing the logger prints is
longer than `MAX_LENGTH`.

morphqdd:
`s[0..32]` is 33 characters, `...` is three more and `s[-32..]` is 32, which makes 68, so the shortening produced something longer than the limit it cuts to. A 63 character string is already 65 with the quotes `inspect` adds, crosses the threshold, and came out at 68.

The cut now goes through `ellipsized`, which the sibling printer `FactAsYaml` already uses and which lands exactly on the limit. The branch that picked the rendering is also written once instead of assigning twice, since the second assignment overwrote the first for every string.

`test_proper_logging` pinned the 68 character rendering, which is the defect written down, so that literal is updated to the one that fits.

All 645 tests stay green.

Closes #892


yegor256:
Merged, thanks! Reusing `ellipsized`, which the gemspec already requires, is better than hand-rolling the arithmetic.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+10520](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #926, opened 2026-09-15
#895: end a string literal at its closing quote

### LINKED ISSUE #895: Text glued to a closing quote is absorbed into the string literal
## What happens

The tokenizer does not end a token at a closing quote (`lib/factbase/syntax.rb:102`), and the
literal is then built by stripping the first and the last character of whatever the token turned
out to be (`syntax.rb:141`, `t[1..-2]`). So a bare word glued to the closing quote is absorbed
into the string, and the quote that was supposed to end it becomes part of the value:

```ruby
Factbase::Syntax.new("(eq foo 'a'bar)").to_term.to_s
# => "(eq foo 'a\\'ba')"

fb.insert.foo = 'aba'
fb.query("(eq foo 'a'bar)").each.to_a.size   # => 0
```

The literal the user wrote was `a`. What the term holds is `a'ba`, and nothing is reported.

The mirror shape is policed: `x'a'` is refused with `Wrong symbol format (x'a')`. So the
tokenizer checks one side of a literal and not the other.

This is not #825 (the `(` is not a separator) and not #826 (a quote of the other kind ends the
literal): here both quotes are of the same kind and correctly paired, and what corrupts the
value is the trailing word.

## What should happen

The token should end at the closing quote, or a token with characters after that quote should be
refused, the way `x'a'` is.

opusstudiohq-max:
Closes #895.

The tokenizer noticed a closing quote only to clear `opener`; it kept accumulating characters into the same token afterwards, and the literal was then built by cutting the first and last character off whatever the token turned out to be. So a word glued to the closing quote was absorbed into the string, and the quote meant to end it became part of the value:

```ruby
Factbase::Syntax.new("(eq foo 'a'bar)").to_term.to_s
# => "(eq foo 'a\'ba')"       the user wrote a; the term holds a'ba, silently
```

The mirror shape was already policed — `x'a'` is refused with *Wrong symbol format*. This refuses the other side too: the closing quote now ends the token where it stands, and the next character has to be a separator, or the query is rejected naming what was glued on.

Escaped quotes take a different branch and are untouched; the existing tests with escaped quotes inside literals (`'Dude\'s Friend'`, `'I\'m \"good\"'`) all still pass.

Verified on my fork first: 646 tests, 0 failures, `rake` green on ubuntu, macos and windows. The new test fails without the change because no error is raised at all.

---


yegor256:
CI is green and the change does what #895 asks for, but I do not want to merge it without a second opinion.

It turns input that the tokenizer used to accept into a hard `ArgumentError`, in a parser that `fbe`, `judges` and `judges-action` all build query strings for programmatically. zerocracy/fbe#965 is an example of a caller that emits `'a\'` from a value ending in a backslash; today that comes out as `String not closed`, and after this change the surrounding text would meet the new `closed` guard instead. The set of characters allowed after a closing quote here is `' '`, `)`, `\n`, `\t`, `\r` and `(` — anything else now raises, including a quote, so `'a''b'` is refused too.

That may well be the intended outcome, but it is a compatibility decision about the query language rather than a local fix, so it needs your call. Assigning to you.

---

PR #933, opened 2026-09-16
#848: pin the pdd action to a commit instead of a branch

### LINKED ISSUE #848: PDD CI runs an unpinned action from the mutable master branch
## The bug
`.github/workflows/pdd.yml` invokes `volodya-lombrozo/pdd-action@master`. The same commit can therefore execute different action code over time, and a branch change in that third-party repository can alter or compromise the check without any change to this repository.

## Expected
Pin the action to an immutable commit SHA (with a version comment), so PDD results are reproducible and reviewable.

opusstudiohq-max:
`volodya-lombrozo/pdd-action@master` resolves at run time, so the same commit of this repository can run different action code on different days, and a push to that branch changes what runs here without anything changing here.

The action is pinned to `69842b56627431c5f232f5ea2dc2fca82f409c54`, which is what `master` points at today and has since 2023-12-18, so the job runs exactly the code it runs now. The comment carries the date rather than a version, because the action publishes no tags or releases to name.

Verified on my fork first: the `pdd` job runs and passes against the pinned commit.

Closes #848

---


opusstudiohq-max:
Closing this until #848 is triaged. It carries no label, so the project has not accepted it as a defect yet, and a pull request opened ahead of triage asks for review work on a problem nobody has agreed to fix — and goes stale at a cost to the reviewers and reporters here, not only to me. Raised by @Thayorns on zerocracy/judges-action#2298, and it applies to my whole queue, not only to that repository. The branch stays on my fork; I will reopen this if and when #848 gets a label.

PR #958, opened 2026-09-16
#942: read every value of a boolean property

### LINKED ISSUE #942: `and`, `or` and `not` look only at the first value of a multi-valued property
## What happens

`Factbase::Boolean#bool?` in `lib/factbase/terms/boolean.rb:23` does `val = val[0] if val.respond_to?(:each)` and throws the rest away. `_values` hands it every value of the property, so `(or flag)` answers false for a fact whose `flag` holds a `true`, only because a `false` was written first.

```ruby
fb.insert.then { |f| f.flag = false; f.flag = true }   # flag => [false, true]
fb.query('(or flag)').each.to_a.size    # => 0
fb.query('(and flag)').each.to_a.size   # => 0
fb.query('(not flag)').each.to_a.size   # => 1
```

Every other family of terms is documented and implemented as "any value matches": `(zero v)` is true if any `v` is zero, `(eq v1 v2)` is true if any `v1` equals any `v2`.

#795 lists `eq`, `not` and `zero` on a single-valued boolean. The multi-valued drop inside `bool?` is not covered there.

## What should happen

`(or p)` should be true if any value of `p` is true, and `(and p)` true only if all of them are, consistent with the "any value" rule used by the rest of the language.

morphqdd:
`Factbase::Boolean#bool?` kept only the first value of the property, so `(or flag)` answered false for a fact whose `flag` held a `true`, just because a `false` was written first. Now `bool?` is true when any value is true, like the rest of the language, and `and` uses a new `every?` that requires all of them.

Closes #942


morphqdd:

PR #959, opened 2026-09-16
#894: print a time as a string, like the other printers

### LINKED ISSUE #894: `Factbase::ToYAML` produces YAML that `YAML.load` refuses to read back
## What happens

`Factbase::ToYAML` dumps the facts with `YAML.dump` (`lib/factbase/to_yaml.rb:31`), which writes
a `Time` as a native YAML object. In Psych 4, `YAML.load` is `safe_load` and `Time` is not
permitted, so the gem's own output is not readable by the default reader:

```ruby
YAML.load(Factbase::ToYAML.new(fb).yaml)
# Psych::DisallowedClass: Tried to load unspecified class: Time
```

`Factbase::Fact` explicitly allows a `Time` as a property value (`fact.rb:58-60`), so this is an
ordinary factbase, not a corner case. `ToJSON` renders a time as an ISO 8601 string on purpose,
which is why its output is plain, and nothing in the rdoc of `ToYAML` warns that its output
needs `permitted_classes: [Time]`.

## What should happen

Either render a time as a string, the way the other printers do, or say in the rdoc that the
output has to be read with `permitted_classes: [Time]`.

morphqdd:
`ToYAML` dumped a `Time` as a native YAML object, which `YAML.load` refuses to read in Psych 4, so the gem could not read its own output. It now prints a time as ISO 8601, the way `ToJSON` and `ToXML` do.

This changes what `test_time_value` expects: the value read back is now the string, not a `Time`. The other option in the issue was to leave the dump alone and document `permitted_classes: [Time]` instead, so say the word if you prefer that one.

Closes #894


morphqdd:

PR #962, opened 2026-09-16
#936: read a string as a decimal number

### LINKED ISSUE #936: `to_integer` reads '010' as octal and '0x1f' as hexadecimal, unlike `to_float` and unlike the tokenizer
## What happens

`lib/factbase/terms/to_integer.rb:32` calls `Integer(value)` with no base. `Kernel#Integer` honours Ruby literal prefixes, so a leading zero means octal and `0x` means hexadecimal. `to_float` in `lib/factbase/terms/to_float.rb:32` uses `Float(value.to_s)`, which has no such rule, and the tokenizer in `lib/factbase/syntax.rb` parses number literals with an explicit base, `Integer(t, 10)`.

The fact `s = '010'`, `h = '0x1f'`:

```
(eq 8 (to_integer s))   => 1 fact matched
(eq 10 (to_integer s))  => 0 facts matched
(eq 10.0 (to_float s))  => 1 fact matched
(eq 31 (to_integer h))  => 1 fact matched
(eq 10 010)             => 1 fact matched
```

So the same three characters mean 10 when written as a literal in a query, 10.0 through `to_float`, and 8 through `to_integer`.

This arrived with the fix for #672, which replaced `Integer(value.to_s)` with `Integer(value)`.

## What should happen

`(to_integer '010')` should be `10`, the same as `(to_float '010')` and the same as the literal `010`. The conversion should pass base 10 explicitly for String input.

morphqdd:
`Integer(value)` honours Ruby literal prefixes, so `'010'` came back as 8 and `'0x1f'` as 31, while `to_float` and the tokenizer both read the same text as decimal. A String now goes through `Integer(value, 10)`; other types are converted as before.

Touches `lib/factbase/terms/to_integer.rb`, which open PR #931 also touches, so this is a draft.

Closes #936


yegor256:
Merged, thanks! Reading a string in base ten makes `to_integer` agree with `to_float` and with the tokenizer.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+11314](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #963, opened 2026-09-16
#937: read the environment variable by the name given

### LINKED ISSUE #937: `env` upcases the variable name, so a lower-case variable is unreachable and a different one is read instead
## What happens

`lib/factbase/terms/env.rb:27` reads `ENV.fetch(n[0].upcase, nil)`. The README says `(env v1 v2)` returns the value of environment variable `v1`. It returns the value of `v1.upcase` instead.

Two consequences. A variable whose real name has a lower-case letter can never be read, and if a variable with the same name in upper case exists, its value is returned silently under the wrong name.

```ruby
ENV['myVar'] = 'mine'
ENV['MYVAR'] = 'other'
ENV.fetch('myVar', nil)                                    # => "mine"
fb.query("(as r (env 'myVar' 'default'))").each.to_a.first # => r => ["other"]
```

With only `myVar` set and no `MYVAR`, the term returns the default, as if the variable were not set at all.

## What should happen

The term should look up the name exactly as given.

morphqdd:
`env` upcased the name before the lookup, so a variable with a lower-case letter was unreachable and a same-name-different-case variable was returned in its place. The name is now used as given.

Touches `lib/factbase/terms/env.rb`, which open PR #910 also touches, so this is a draft.

Closes #937


yegor256:
The fix matches what #937 asks for and the tests are convincing, but I am holding off on merging it. Dropping the `upcase` changes the meaning of every existing `(env ...)` term: a query written as `(env "github_token" ...)` reads `GITHUB_TOKEN` today and would read nothing after this change. Nothing in this repository tells us whether such queries exist downstream, in `judges` or `judges-action`, and the failure would be silent — the default operand would be taken instead of the variable.

Could you confirm that no caller depends on the upcasing, or that this should be released as a breaking change? Assigning to you for that call.

---

PR #978, opened 2026-09-17
#945: wrap a literal into an array instead of decomposing it with Kernel#Array

### LINKED ISSUE #945: `Factbase::IndexedEq` turns a `Time` literal into ten integers, so `(eq when <time>)` finds nothing
## What happens

`_resolve` in `lib/factbase/indexed/indexed_eq.rb:50` ends with `return Array(operand) unless operand.is_a?(Symbol)`. `Kernel#Array` tries `to_ary`, then `to_a`. `Time#to_a` is public and returns the ten-element decomposition `[sec, min, hour, day, mon, year, wday, yday, isdst, zone]`.

So for a `Time` literal, `_resolve` hands back ten integers instead of one time. `predict` looks each of them up in `entry[:facts]`, a hash keyed by the stored `Time` objects, finds nothing, and returns an empty candidate list. `Factbase::Query#each` iterates only that list, so the query yields zero facts. `_scalar?` explicitly admits `Time`, so this is the normal path for every time comparison under `IndexedFactbase`.

```
plain  (eq when 2024-03-01T00:00:00Z) -> 2 facts
idx    (eq when 2024-03-01T00:00:00Z) -> 0 facts
Array(Time) = [0, 0, 0, 1, 3, 2024, 5, 61, false, "UTC"]
```

## What should happen

`(eq when <time>)` under `IndexedFactbase` must return the same facts as under a plain `Factbase`.

morphqdd:
`Kernel#Array` calls `Time#to_a`, which decomposes a time into ten integers,
so the index looked up ten numbers instead of one time and found nothing.
A literal is now wrapped into a one-element array.

Closes #945


morphqdd:

PR #979, opened 2026-09-17
#947: index a number by its exact rational, so 1 and 1.0 land on one key

### LINKED ISSUE #947: `Factbase::IndexedEq` looks values up with `Hash#[]`, so `(eq a 1.0)` never matches the Integer `1`
## What happens

`_feed` (`lib/factbase/indexed/indexed_eq.rb:27`) keys the inverted index by the raw stored value, and `predict` (`:42`) reads it back with `entry[:facts][k]`. Hash lookup uses `eql?` and `hash`. `1.eql?(1.0)` is false while `1 == 1.0` is true.

The plain `eq` term compares with `==`, so it matches across numeric types. The index cannot.

```
(eq a 1.0)   plain=["i"]  idx=[]
(eq a 2)     plain=["f"]  idx=[]
(eq a 1)     plain=["i"]  idx=["i"]
(eq a 2.0)   plain=["f"]  idx=["f"]
```

(`i` holds the Integer `1`, `f` holds the Float `2.0`.)

## What should happen

`(eq a 1.0)` must return the same facts as on a plain `Factbase`, whichever numeric type the value was stored as.

morphqdd:
A Hash tells 1 from 1.0, while the plain `eq` term compares with `==` and
does not, so `(eq a 1.0)` found nothing under the index. Numbers are now
kept in the index as exact rationals.

Closes #947


morphqdd:

PR #981, opened 2026-09-17
#957: refuse a non-positive timeout in Factbase::Impatient

### LINKED ISSUE #957: `Factbase::Impatient` with `timeout: 0` silently runs without any timeout
## What happens

The constructor at `lib/factbase/impatient.rb:23` accepts any numeric timeout through `Float(timeout)`, and `impatient` at `:87` passes it straight to `Timeout.timeout(@timeout, &)`.

Ruby's `Timeout.timeout` treats `0`, like `nil`, as no limit, and simply yields. So `Factbase::Impatient.new(fb, timeout: 0)`, the most natural way to ask for "give up immediately", produces a decorator that never interrupts anything while still reporting itself as impatient.

```
Timeout.timeout(0) did NOT time out
```

There is no validation and no warning, so a mis-configured or zeroed-out timeout turns the whole mechanism off with no signal.

## What should happen

A non-positive timeout must be refused in the constructor with an `ArgumentError`, rather than silently disabling the only thing the class exists to do.

### LINKED ISSUE #999: `Factbase::Impatient` with a negative timeout makes every query raise a raw `ArgumentError`
## What happens

`Factbase::Impatient.new` takes any number (`lib/factbase/impatient.rb:23`):

```ruby
@timeout = Float(timeout)
```

and hands it to `Timeout.timeout` (`:87`). Ruby refuses a negative number there, with
`ArgumentError: Timeout sec must be a non-negative number`, and `impatient` rescues only
`Timeout::Error`, so the error comes out raw from every call the decorator makes, for as long as
the object lives:

```ruby
imp = Factbase::Impatient.new(Factbase.new, timeout: -1)
imp.query('(exists foo)').each.to_a
```

```
ArgumentError: Timeout sec must be a non-negative number
```

The factbase is not slow and the query is not broken, but nothing works any more, and the message
says nothing about the factbase or the query.

This is the other half of #957, which is about `timeout: 0` turning the timeout off in silence.
Here the same absent validation makes the whole decorator unusable.

## What should happen

A negative timeout should be refused in the constructor, the way a non-positive one should be,
so the mistake is reported once, where it was made.

morphqdd:
`Timeout.timeout` yields without any limit for `0`, and refuses a negative number with an
`ArgumentError` the decorator does not rescue, so the constructor now takes only a positive
timeout.

Closes #957
Closes #999


morphqdd:

PR #982, opened 2026-09-17
#956: log the number of the fact that was actually inserted

### LINKED ISSUE #956: `Factbase::Logged#insert` logs the size taken before the insert, so the first fact is reported as `#0`
## What happens

`lib/factbase/logged.rb:39` builds and sends the message, and `@origin.insert` runs on line 40. `@origin.size` is therefore the count before the new fact exists.

```
Inserted new fact #0 in 3us
Inserted new fact #1 in 1us
Inserted new fact #2 in 1us
size after three inserts = 3
```

The fact the log calls `#0` is the first of three, and no fact is ever logged under the number the factbase would give it. Because the line is emitted first, an `insert` that raises still leaves a line claiming the fact was inserted.

## What should happen

The number in the message must identify the fact that was actually inserted, and the line must be written after the insert succeeds.

morphqdd:
The line was built before `@origin.insert` ran, so the first fact was
reported as `#0` and a failed insert still logged a line. The message is
now written after the insert succeeds.

Closes #956


morphqdd:

PR #983, opened 2026-09-17
#954: do not cache a term that reads the environment

### LINKED ISSUE #954: `Factbase::CachedTerm` caches `(env ...)`, so a changed environment variable is never seen
## What happens

`lib/factbase/cached/cached_term.rb:18-24` caches under `[maps.object_id, to_s]` for any term that is `static?` and not `abstract?`, excluding only `:head` and `:unique`.

`(env "NAME" "default")` has two String operands, so it is static and not abstract, and every node of the tree is redressed by `CachedFactbase#to_term`, so the `env` node itself is cached. The entry is keyed on the term text and the maps identity only, nothing about the environment. `CachedQuery#invalidate_if_dirty!` clears the cache only after an insert or a write, so as long as the factbase is untouched the first value of the variable is returned forever.

```
first            -> 1
after ENV change -> 1   (a plain factbase answers 0)
```

`env` is the one term whose purpose is to read state living outside the factbase, which is exactly the state the cache key cannot see.

## What should happen

`(env ...)` must not be cached beyond a single query evaluation, so a later query observes the current environment.

morphqdd:
`(env ...)` reads state that lives outside the factbase, which the cache key
cannot see, so a changed variable was never noticed. A term that has `env`
anywhere in it is no longer cached, neither at term level nor at query level.

Closes #954


morphqdd:

PR #984, opened 2026-09-17
#955: read a property from the fact when the fact already holds it

### LINKED ISSUE #955: `Factbase::Accum` answers `f.foo` with the value just written, while the same fact answers with its first value
## What happens

Every fact yielded by `Factbase::Query#each` is a `Factbase::Accum` with `pass = true`. A write goes to both `@props` and the underlying fact (`lib/factbase/accum.rb:37-39`), but a read takes the branch at `:56-60`: `elsif @props[k].nil? ... else @props[k][0]`. So once anything has been written to `foo` through the accumulator, the reader gets the first value written through the accumulator, and the values the fact already held become invisible. `Factbase::Fact#others` returns `v[0]`, the first value of the whole set.

For a fact that already holds `bar = 1`, writing `bar = 5` inside a query block makes `f.bar` answer `5` there and `1` everywhere else, on the same underlying fact.

```
before write: q.bar=1  q['bar']=[1]
after  write: q.bar=5  q['bar']=[1, 5]
plain fact:   g.bar=1  g['bar']=[1, 5]
```

`f['bar']` stays consistent inside and out, so the two accessors disagree only for `f.bar`.

#827 and the closed #477 circle the `f.foo` versus `f['foo']` distinction, but neither describes this read-after-write divergence for the same accessor.

## What should happen

`f.foo` read through the accumulator must return the same value the fact returns, the first value of the merged set.

morphqdd:
A write through the accumulator went to the fact as well, but a later read
of `f.foo` answered from the accumulator and hid the values the fact already
held. A property the fact holds is now read from the fact, so `f.foo` and
`f['foo']` agree, inside a query block and outside it.

Closes #955


morphqdd:

PR #985, opened 2026-09-17
#948: merge the operands of or by identity, not by content

### LINKED ISSUE #948: `Factbase::IndexedOr` merges with `Array#|`, so two facts with identical content collapse into one
## What happens

`predict` in `lib/factbase/indexed/indexed_or.rb:23` accumulates operand results with `r |= n.to_a`. `Array#|` deduplicates by `hash` and `eql?`, and facts at this level are plain Hashes, so two distinct facts holding identical properties are `eql?` and one of them is dropped from the candidate list before the term is ever evaluated.

Every sibling in the directory that faces this deduplicates with `uniq(&:object_id)` instead: `indexed_eq.rb:28`, `indexed_and.rb:45`.

```
plain (or (eq a 1) (eq a 99)) -> 2
idx   (or (eq a 1) (eq a 99)) -> 1
```

`delete!` inherits the same candidate list, so a duplicate fact also survives a delete that should remove it.

This is not #844, which is about one fact carrying a repeated value. Here two separate facts are lost.

## What should happen

`(or ...)` must yield every matching fact, including facts whose property sets happen to be equal.

morphqdd:
`Array#|` compares facts by content, so two separate facts holding the same
properties collapsed into one before the term was evaluated. The operands
are now merged by identity, the way `eq` and `and` already do it.

Closes #948


morphqdd:

PR #986, opened 2026-09-17
#949: yield the facts of or in insertion order

### LINKED ISSUE #949: `Factbase::IndexedOr` returns facts in operand order, not in insertion order
## What happens

`predict` in `lib/factbase/indexed/indexed_or.rb:16-23` walks the operands one at a time and appends each operand's hits to `r`. The candidate list is therefore grouped by which operand matched, and `Factbase::Query#each` yields in exactly that order.

So `(or (eq a 1) (eq b 2))` returns every `a = 1` fact before every `b = 2` fact, regardless of where they sit in the factbase. The plain `Factbase` returns them in insertion order.

```
plain order -> ["first", "second"]
idx   order -> ["second", "first"]

(or (gt b 55) (lt b 2))  plain first three: b=0, b=1, b=56
                         idx   first three: b=56, b=57, b=58
```

#785 and #845 fixed this class of defect for `gt`, `gte`, `lt` and `lte`. `or` was never touched.

## What should happen

An `or` query must yield facts in insertion order, like every other query.

morphqdd:
The operands were walked one by one and their hits appended, so the facts
came out grouped by which operand matched. They are now selected from the
factbase in the order it holds them.

Closes #949


morphqdd:

PR #987, opened 2026-09-17
#950: index not only when its sub-term predicts exactly

### LINKED ISSUE #950: `Factbase::IndexedNot` treats the sub-term's prediction as exact, so `(not (and ...))` drops matching facts
## What happens

`predict` is a filter that is allowed to over-approximate: returning a superset is always safe, because `Factbase::Query#each` still evaluates the real term on every candidate. Several terms rely on that. `IndexedOr` returns the whole `maps` as soon as the union exceeds a quarter of the factbase (`indexed_or.rb:24`), `IndexedGte`, `IndexedGt`, `IndexedLt` and `IndexedLte` return the whole `maps` when a parameter is missing, and `IndexedAnd` stops intersecting early (`indexed_and.rb:62-63`).

`IndexedNot#_feed` (`lib/factbase/indexed/indexed_not.rb:41-43`) builds `yes_set` from that prediction and then returns `facts.reject { |m| yes_set.include?(m) }` as the final answer. It never evaluates the negation. Every fact the sub-term over-approximated into `yes_set` is removed from the result even though it does not match. Unlike the positive terms, `not` under-approximates here, and that is not recoverable by the later evaluation.

```
(not (or (eq a 1) (eq b 3)))   plain=50  idx=0
(not (gte b $missing))         plain=100 idx=0

randomized sweep over 40 facts:
(not (and (exists c) (absent a)))  plain=32 idx=10
(not (and (absent a) (lt a 2)))    plain=40 idx=29
(not (and (eq a 1) (absent a)))    plain=40 idx=32
(not (and (one a) (eq c 2)))       plain=34 idx=20
seventeen distinct (not (and ...)) shapes mismatched in one run
```

## What should happen

`not` must either decline to index whenever its sub-term's prediction is not exact, or return a superset and let the full evaluation decide.

morphqdd:
A prediction is allowed to answer with more facts than match, because the
query evaluates the term afterwards. For `not` that is not recoverable: the
extra facts were subtracted from the answer and never evaluated. `not` now
indexes only when its sub-term is an `eq` over a property and a value, and
otherwise leaves the work to the full scan.

Closes #950


morphqdd:

PR #988, opened 2026-09-17
#952: decline to index a dollar-prefixed operand in exists and one

### LINKED ISSUE #952: `Factbase::IndexedExists` and `Factbase::IndexedOne` read `$x` as a property name, so the query returns nothing
## What happens

`lib/factbase/indexed/indexed_exists.rb:19` does `operand = @term.operands.first.to_s` with no type check. `lib/factbase/indexed/indexed_one.rb:15-16` guards only with `operand.is_a?(Symbol)`, which a parameter reference satisfies, because `$x` parses to the symbol `:$x`.

Both then build the literal property name `"$x"`, find no fact carrying it, and return an empty candidate list, so the query yields nothing. The plain terms resolve the parameter through `Tee` and match.

```
(exists $x)   plain=["p", "q", "r"]  idx=[]
(one $x)      plain=["p", "q", "r"]  idx=[]
(absent $x)   plain=[]               idx=[]
```

#784 fixed the sub-term case for `IndexedOne` by adding the Symbol guard; a `$`-prefixed symbol slips straight through it, and `IndexedExists` never got a guard at all.

## What should happen

A `$`-prefixed operand must make these terms decline to index, so the full scan resolves the parameter.

morphqdd:
`$x` parses to the symbol `:$x`, so both terms built the literal property
name `"$x"`, found no fact carrying it and returned nothing. A dollar
prefixed operand now makes them decline, so the full scan resolves the
parameter.

Closes #952


morphqdd:

PR #989, opened 2026-09-17
#951: yield the facts of eq and and in insertion order

### LINKED ISSUE #951: `Factbase::IndexedEq` with a multi-valued parameter returns facts grouped by parameter value, not in insertion order
## What happens

`lib/factbase/indexed/indexed_eq.rb:27` reads `matches = keys.flat_map { |k| entry[:facts][k] || [] }`. It iterates the parameter's values and concatenates the bucket for each one. With `params = { 'x' => [2, 1] }`, every fact with `a = 2` is emitted before every fact with `a = 1`. The plain `Factbase` scans `@maps` once and preserves insertion order.

`IndexedAnd`'s `multi_and_eq` branch (`lib/factbase/indexed/indexed_and.rb:45`) has the same `flat_map` shape and reorders the same way.

```
order plain = ["p", "q", "r"]
order idx   = ["q", "p", "r"]

reordered: (eq a $p)
reordered: (and (eq b "x") (eq a $p))
reordered: (and (eq a $p) (eq c 2))
```

Same class as #785 and #845, for `eq` and `and` rather than the range terms.

## What should happen

The result order must not depend on the order of the values inside a query parameter.

morphqdd:
A parameter holding several values was walked value by value and the hits
appended, so the facts came out grouped by value instead of in insertion
order. Both `eq` and the `multi_and_eq` branch of `and` now answer in the
order the factbase holds the facts.

Closes #951


morphqdd:

PR #990, opened 2026-09-17
#953: check a rule against the facts the factbase holds

### LINKED ISSUE #953: `Factbase::Rules::Check#it` evaluates the rule against an empty `maps`, so any rule containing `agg` always fails
## What happens

`lib/factbase/rules.rb:131` reads `Factbase::Syntax.new(@expr).to_term.evaluate(fact, [], fb)`. The second argument, the list of all facts, is hard-coded to `[]`.

`Factbase::Agg#evaluate` builds its sub-query as `fb.query(selector, maps)`, and `Factbase#query` honours an explicitly supplied `maps`, so the aggregate runs over an empty array no matter what the factbase holds. A rule such as `(gt (agg (always) (count)) 0)` sees a count of `0` and raises on every property write, while the same expression run as a query returns `2`.

```
base size = 2
direct query, agg count = 2
rule (agg count > 0) FAILED:
  The fact doesn't match the "(gt (agg (always) (count)) 0)" rule: [ ] + {"foo" => [9]}
```

Any rule that reasons about other facts, such as uniqueness of a key, a referenced parent existing, or a bounded total, is unusable and rejects everything.

## What should happen

`Check#it` must pass the factbase's real facts as `maps`, so a rule containing `agg` sees the same data a query would.

morphqdd:
The rule was evaluated against an empty list of facts, so `agg` inside a
rule counted zero and rejected every write. The check now sees the facts
the factbase holds.

Closes #953


morphqdd:

PR #991, opened 2026-09-17
#944: refuse a non-term operator while the AST is built

### LINKED ISSUE #944: The parser accepts any token as the operator, so `(42 foo)` builds a term and fails only at evaluation time
## What happens

`to_ast` in `lib/factbase/syntax.rb:79` takes `op = tokens[at]` without checking that it is a Symbol. An Integer, a string literal or a nested opening bracket becomes the term name, the query parses cleanly, and the failure only shows up per fact inside `__send__`.

```
"(42 foo)"      parses into a Factbase::Term
                then, at query time:
                !! "42 is not a symbol nor a string" at (42 foo) at .../term.rb:208

"((eq foo 1))"  !! Factbase::Syntax::Broken: Too many terms (6 != 7)
```

The second message is a token count, which says nothing about the doubled bracket that caused it. The first message never mentions the query at all until the term is evaluated against a fact.

This is not #825, which is about `(` not acting as a token separator.

## What should happen

A non-Symbol operator should be refused while the AST is built, with a message naming the offending token and the query, and a nested opening bracket should be reported as such rather than as a term count.

morphqdd:
Any token was taken as the operator, so `(42 foo)` built a term and failed
only when it was evaluated, and a doubled bracket was reported as a token
count. The operator is now checked while the AST is built, and the message
names what was found instead.

Closes #944


morphqdd:

PR #995, opened 2026-09-17
fix(#964): refuse a non-string value in matches instead of converting it

### LINKED ISSUE #964: `matches` converts a non-string property while `contains`, `starts_with` and `ends_with` refuse it
## What happens

The README lists four predicates side by side, on consecutive lines, with the same shape:

```
* `(matches v s)`     — if any `v` matches the `s` regular expression
* `(contains v s)`    — if any value of `v` contains any value of `s` (substring match)
* `(starts_with v s)` — if any `v` starts with `s`
* `(ends_with v s)`   — if any `v` ends with `s`
```

Three of them refuse a property that is not a string. The fourth converts it and answers.

`Contains`, `StartsWith` and `EndsWith` each delegate to `Factbase::Compare` (`terms/contains.rb:17` and siblings), and `Compare#_compare` turns Ruby's refusal into the message #791 introduced. `Matches` does not go through `Compare`; it ends with a bare conversion at `terms/matches.rb:33`:

```ruby
str[0].to_s.match?(regexp(re[0]))
```

## Reproduction

One fact carrying an `Integer`, a `Float` and a `Time`, queried with each of the four:

```ruby
fb = Factbase.new
f = fb.insert
f.num = 42
f.flo = 3.5
f.tim = Time.parse('2024-03-05T06:07:08Z')
```

```
(matches num "42")         => 1                 <- matched
(contains num "42")        => RuntimeError: Cannot compare 42 (Integer) with "42" (String) using (compare include?)
(starts_with num "42")     => RuntimeError: Cannot compare 42 (Integer) with "42" (String) using (compare start_with?)
(ends_with num "42")       => RuntimeError: Cannot compare 42 (Integer) with "42" (String) using (compare end_with?)

(matches flo "3.5")        => 1                 <- matched
(contains flo "3.5")       => RuntimeError: Cannot compare 3.5 (Float) with "3.5" (String) ...

(matches tim "2024")       => 1                 <- matched
(contains tim "2024")      => RuntimeError: Cannot compare 2024-03-05 06:07:08 UTC (Time) with "2024" (String) ...
```

All four agree on a real string, so the difference is only in how they treat a value that is not one.

## Why the `Time` case is the one that bites

`(matches tim "2024")` does not ask a question about a moment. It runs the regexp over whatever `Time#to_s` produced — here `2024-03-05 06:07:08 UTC`. A query written as `(matches when "^2024-03")` looks like a date filter and is really a substring test against one particular rendering: it reads the space separator, the `UTC` suffix and the absence of a `T`, none of which the query author chose. Change the stored value to a `Time` in another zone and the same query stops matching. `contains` refuses to answer that question at all, which is the safer answer of the two.

## This is not #829

#829 is about a different axis — `matches` refusing a *multi-valued* property that the other three accept — and PR #853 is open for it. That patch replaces `str[0].to_s.match?` with `str.any? { |s| s.to_s.match?(rx) }`, so the `to_s` stays exactly where it is and every line of the reproduction above behaves the same after it lands. The two do not overlap.

## What should happen

Either `matches` refuses a non-string the way its three siblings do, by going through `Factbase::Compare` like they do, or the four stop being documented as one family. I lean towards the first — a silent `to_s` on a `Time` answers a question nobody asked — but that is a decision about the language, so I am leaving it here rather than picking one.

Thayorns:
`matches` used to call `to_s` on the property before running the regexp, so an Integer, a Float, a Time or a boolean was silently turned into one particular string rendering and matched against it. Its siblings `contains`, `starts_with` and `ends_with` refuse such a value through `Factbase::Compare`. Now `matches` refuses it too, raising a `RuntimeError` that names the value, its class and the regexp.

I did not route `matches` through `Factbase::Compare` itself. Doing so would also make it accept multi-valued properties, which is what #829 and the open #853 are about, and it would drop the compiled regexp cache and the invalid regexp message. The change stays on the non-string axis only.

The old assertion that `42` quietly fails to match `[a-z]+` described exactly the conversion this issue reports, so it is replaced by tests with random seeded inputs for each non-string type and one through a full query.

Closes #964


Thayorns:
@yegor256 happy to clarify why this stays off the Compare path, if that part looks odd.

Thayorns:
@yegor256, fixed, have a look

PR #1012, opened 2026-09-18
#998: check the whole property name, not one line of it

### LINKED ISSUE #998: `Factbase::Fact` checks a property name with line anchors, so a name with a newline in it is accepted
## What happens

`lib/factbase/fact.rb:53` guards a property name with

```ruby
raise(ArgumentError, "Invalid prop name '#{kk}'") unless kk.match?(/^[a-z_][_a-zA-Z0-9]*$/)
```

In Ruby `^` and `$` match at a line boundary, not at the boundary of the string, so a name only
needs one line that looks like an identifier for the whole name to pass. A name that carries a
newline therefore gets in, with whatever the other lines hold:

```ruby
f = Factbase::Fact.new({})
f.public_send(:"weird\nProp!!!=", 42)
f.all_properties
```

```
["weird\nProp!!!"]
```

A plain bad name without a newline is refused, as it should be, so the guard looks like it works
until a newline shows up.

## What should happen

The name should be checked against the whole string, with `\A` and `\z`, so that a multi-line
name is refused the same way `Bad!Name` is.

morphqdd:
`^` and `$` match at a line boundary, so a name that holds a newline passed as long as one of its
lines looked like an identifier. The check now covers the whole string.

Closes #998


morphqdd:

yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12334](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1013, opened 2026-09-18
#1002: report an infinite float the way every other bad conversion is reported

### LINKED ISSUE #1002: `to_integer` of an infinite float escapes as a raw `FloatDomainError`
## What happens

`lib/factbase/terms/to_integer.rb:22-27` wraps the conversion:

```ruby
Integer(value)
rescue ArgumentError, TypeError => e
  raise(RuntimeError, "Cannot convert #{value.inspect} to Integer in #{self}: #{e.message}")
```

`Integer(Float::INFINITY)` raises `FloatDomainError`, which is a `RangeError` and not in that
list, so it escapes unwrapped. An infinite float is reachable from a query, because a literal
like `1.0e400` parses to `Infinity`:

```ruby
Factbase::ToInteger.new([1.0 / 0]).evaluate(nil, nil, nil)
```

```
FloatDomainError: Infinity
```

The message names neither the term nor the query, while every other bad conversion in the same
file is reported with both.

## What should happen

An infinite or NaN value should be reported the way every other bad conversion is, with the term
and the value in the message.

morphqdd:
`Integer(Float::INFINITY)` raises `FloatDomainError`, which is a `RangeError` and was not in the
rescue list, so it escaped without the term or the value in the message.

Closes #1002


morphqdd:

PR #1014, opened 2026-09-18
#1004: remember a number the way eq compares it

### LINKED ISSUE #1004: `unique` tells `1` and `1.0` apart, while `eq` calls them the same value
## What happens

`lib/factbase/terms/unique.rb:25-26` remembers what it has seen in a `Set`:

```ruby
Enumerator.product(*vv).to_a.all? { |t| @seen.add?(t) }
```

`Set` tells its members apart with `eql?`, which says that `1` and `1.0` are two different
things. `Factbase::Compare`, which `eq` is built on, compares with `==`, which says they are the
same thing. So the two terms disagree about what one value is:

```ruby
f1.x = 1
f2.x = 1.0
Factbase::Term.new(:eq, [:x, 1.0]).evaluate(f1, [f1], fb)   # true

u = Factbase::Term.new(:unique, [:x])
u.evaluate(f1, [f1, f2], fb)   # true
u.evaluate(f2, [f1, f2], fb)   # true again
```

`(unique x)` is meant to let one fact per value through, and here it lets both through, for a
value `eq` calls the same.

## What should happen

`unique` should tell values apart the way `eq` does, so an Integer and the equal Float count as
one value.

morphqdd:
A `Set` tells `1` and `1.0` apart, while `eq` calls them one value, so `(unique x)` let a Float
duplicate of an Integer through. A number is now remembered as an exact rational.

Closes #1004


morphqdd:

PR #1015, opened 2026-09-18
#1008: guard the feed against an array that shrank

### LINKED ISSUE #1008: `Factbase::IndexedExists` has no shrink guard, so a second query over a shortened array crashes
## What happens

Every indexed term guards its incremental feed against an array that shrank, for example
`indexed_absent.rb:25`, `indexed_eq.rb:39`, `indexed_one.rb:27`, `indexed_gt.rb:35`:

```ruby
return unless entry[:count] < facts.size
```

`lib/factbase/indexed/indexed_exists.rb:29-34` has no such line and slices straight away:

```ruby
facts[entry[:count]..].each do |m|
```

`query(term, maps)` takes the array from the caller, and the entry is kept under the array's
identity. If the caller shrinks its own array and queries again, the slice is past the end, which
in Ruby is `nil`, not an empty array:

```ruby
maps = []
5.times { |i| maps << { 'x' => [i] } }
ifb.query('(exists x)', maps).each(ifb, {}).to_a   # 5 facts, entry[:count] = 5
maps.pop(3)
ifb.query('(exists x)', maps).each(ifb, {}).to_a
```

```
NoMethodError: undefined method 'each' for nil
```

The same sequence with `(absent x)`, `(one x)` or `(eq x 1)` answers normally.

## What should happen

`Factbase::IndexedExists` should carry the same guard its siblings carry, so a shrunk array is
answered rather than crashed on.

morphqdd:
Every sibling indexed term returns early when the array is no longer than what was already fed,
because slicing past the end answers `nil` and not an empty array. `Factbase::IndexedExists` was
the one without that line, so a second query over a shortened array crashed.

Closes #1008


morphqdd:

PR #1016, opened 2026-09-18
#1009: report a query too deep to parse as Broken

### LINKED ISSUE #1009: A deeply nested query escapes as `SystemStackError`, not as `Factbase::Syntax::Broken`
## What happens

`lib/factbase/syntax.rb:43-47` promises one error for a query it cannot read:

```ruby
rescue StandardError => e
  err = "#{e.message} (#{Backtrace.new(e)}) in \"#{@query}\""
  raise(Broken, err)
```

`to_ast` (`:75-94`) calls itself once per bracket with no limit, so a deeply nested query runs out
of native stack. `SystemStackError` descends from `Exception`, not from `StandardError`, so it
goes straight past that rescue:

```ruby
depth = 50_000
q = ('(and ' * depth) + '(eq foo 1)' + (')' * depth)
Factbase::Syntax.new(q).to_term
```

```
UNCAUGHT SystemStackError
```

A caller that reads a query from a file or from a user, and rescues `Factbase::Syntax::Broken`
because that is what the class documents, dies instead.

## What should happen

A query too deep to parse should be reported as `Factbase::Syntax::Broken`, the way every other
unreadable query is.

morphqdd:
`SystemStackError` descends from `Exception`, not from `StandardError`, so a query nested deeply
enough to exhaust the stack went past the rescue that turns every other unreadable query into
`Factbase::Syntax::Broken`.

Closes #1009


morphqdd:

PR #1018, opened 2026-09-18
#1005: search with the loosest of the values bound to the parameter

### LINKED ISSUE #1005: `Factbase::IndexedGt` and its three siblings keep only the first value of a multi-valued parameter, so the query finds nothing
## What happens

`lib/factbase/indexed/indexed_gt.rb:17`, and the same line in `indexed_gte.rb`, `indexed_lt.rb`
and `indexed_lte.rb`, resolve the right operand like this:

```ruby
target = op2.is_a?(Symbol) ? params[op2.to_s]&.first : op2
```

A parameter can hold several values, and `Factbase::Compare` treats them as alternatives:
`rights.any? { |r| _compare(l, r) }`. The index keeps only the first one, so it narrows the
candidate list with a bound the plain term would never have used alone, and `Factbase::Query#each`
iterates only what `predict` returned:

```ruby
5.times { |i| fb.insert.x = i }
fb.query('(gt x $t)').each(fb, { 't' => [10, 1] }).to_a.map(&:x)
ifb.query('(gt x $t)').each(ifb, { 't' => [10, 1] }).to_a.map(&:x)
```

```
plain:   [2, 3, 4]
indexed: []
```

Nothing is raised, the answer is simply smaller.

## What should happen

`(gt x $t)` under `Factbase::IndexedFactbase` must return the same facts as under a plain
`Factbase`, so the index should take the loosest of the parameter's values, the smallest one for
`gt` and `gte` and the largest for `lt` and `lte`.

morphqdd:
`Factbase::Compare` keeps a fact when it compares well against any one of the values bound to the
parameter, and the index kept only the first of them, so it narrowed the candidates with a bound
the term would never have used alone and the query answered less than the plain factbase.

Closes #1005


morphqdd:

PR #1019, opened 2026-09-18
#1006: compare a time the way Factbase::Compare compares it

### LINKED ISSUE #1006: The indexed range terms compare a `Time` at full precision, while `Factbase::Compare` floors it, so a fact near a second boundary is lost
## What happens

`Factbase::Compare` floors a `Time` on both sides before it compares:

```ruby
l = l.floor if l.is_a?(Time)
r = r.floor if r.is_a?(Time)
```

The indexed range terms do not. `lib/factbase/indexed/indexed_gte.rb:34-42` stores the value as it
is and `:52-58` searches it as it is, at full precision, so the two disagree for two times inside
one second, and the fact is gone from the candidate list before the term is evaluated:

```ruby
fact = Time.parse('2024-01-01T00:00:01.9Z')
target = Time.parse('2024-01-01T00:00:01.99Z')
fb.insert.when = fact
fb.query('(gte when $t)').each(fb, { 't' => [target] }).to_a.size
ifb.query('(gte when $t)').each(ifb, { 't' => [target] }).to_a.size
```

```
plain:   1
indexed: 0
```

The same shape is in `indexed_gt.rb`, `indexed_lt.rb` and `indexed_lte.rb`. A "since the last run"
filter is exactly this query.

## What should happen

The index should floor a `Time` the way `Factbase::Compare` floors it, so the prediction is never
smaller than the real answer.

morphqdd:
`Factbase::Compare` floors a `Time` on both sides before it compares, and the index compared at
full precision, so for two times inside one second the two disagreed about the order and the fact
was gone from the candidates before the term was ever evaluated. A "since the last run" filter is
exactly this query.

Closes #1006


morphqdd:

PR #1021, opened 2026-09-18
#994: pin the pdd and xcop actions to a release or a commit

### LINKED ISSUE #994: Factbase workflows execute unpinned external actions
The workflow files .github/workflows/pdd.yml and .github/workflows/xcop.yml reference third-party actions by a moving branch such as `master` instead of an immutable release or commit SHA.\n\n## Reproduction\n\nCheck out any fixed commit of this repository and run the referenced workflow again after the upstream action branch changes. The action code, and therefore the result of the check, can change even though this repository has not changed.\n\n## Expected behavior\n\nPin every third-party action to an immutable version or commit SHA.\n\n## Impact\n\nHistorical CI runs cannot be reproduced reliably, and an unrelated upstream action change can break a green commit without a local change.

Thayorns:
The pdd and xcop workflows ran `volodya-lombrozo/pdd-action@master` and `g4s8/xcop-action@master`, so the code behind those two checks could change under a commit that never changed itself. xcop now runs its `v1.3` release, which is the commit its master points at today. pdd-action has no tags or releases at all, so it is pinned to the commit its master points at today, 69842b5.

A new `test/test_workflows.rb` reads every workflow and fails when a `uses:` reference ends in anything other than a version tag or a full commit SHA. It failed on exactly these two references before the change.

The test still accepts major tags like `actions/checkout@v7` and `ruby/setup-ruby@v1`, which the other workflows use and Renovate keeps current. Those tags move with every upstream release, so full immutability would mean SHAs for them too. I left them alone because the issue is about the branch references.

Closes #994


Thayorns:
@yegor256 happy to explain why pdd is pinned to a commit rather than a tag, if that part looks odd.

Thayorns:
@yegor256 pushed the changes, please review

PR #1022, opened 2026-09-18
fix(#996): pin that parameterized queries never share a cached result

### LINKED ISSUE #996: CachedQuery reuses one parameterized result for every each call
## What happens

A non-abstract query with a parameter is cached by the query text only in `CachedQuery#facts`. The parameter values are not part of the cache key, so the first call permanently supplies the rows for later calls with different parameters.

## Reproduction

```ruby
fb = Factbase::CachedFactbase.new(Factbase.new)
[1, 2].each { |value| fb.insert.value = value }

query = fb.query('(eq value $wanted)')
first = query.each(fb, wanted: [1]).to_a
second = query.each(fb, wanted: [2]).to_a

p first.map { |fact| fact.value }  # => [[1]]
p second.map { |fact| fact.value } # => [[1]], expected [[2]]
```

The same query against a plain `Factbase` returns the expected row for each parameter.

## Why

`lib/factbase/cached/cached_query.rb` builds the iteration key as `"each #{@origin}"` and stores `@origin.each(fb, params).to_a` under it. Unlike `one`, which includes `params` in its key, `facts` ignores the parameters entirely.

## Expected result

The cache key for `each` must include the parameter values, or parameterized queries must bypass this cache. Different parameter sets must never share a cached result.

Thayorns:
The reproduction in #996 does not fail on master. `CachedFactbase#query` builds the `CachedQuery` with `cacheable: !term.abstract?`, and a term can only read a parameter through a `$` symbol, which makes it abstract. So `(eq value $wanted)` never reaches the cache: both `each` and `one` go straight to the origin query and return the right rows for every parameter. A query that takes a parameter and is still not abstract cannot be written, which is the case the issue describes.

This pull request changes no production code. It adds three tests to `test_cached_query.rb` that pin the guard, so the cache cannot start sharing results between parameter sets without someone noticing: a parameter given to `each`, a parameter hidden inside an `agg`, and a parameter given to `one`. Each test warms the cache with one random value and then asks for another. All three pass on master and all three fail once the guard is replaced with `cacheable: true`.

Closes #996


Thayorns:
@yegor256 happy to explain why the issue does not reproduce, or to close this and #996 if you would rather not keep the tests.

Thayorns:
@yegor256 when you have a minute, please check

Thayorns:
@yegor256 updated, take a look

PR #1032, opened 2026-09-19
#1010: a number with an exponent and no dot is a float

### LINKED ISSUE #1010: A number written as `1e10` is refused, because the float pattern requires a dot
## What happens

`lib/factbase/syntax.rb:146` recognises a float only with a dot in it:

```ruby
elsif t.match?(/^(\+|-)?[0-9]+\.[0-9]+(e(\+|-)[0-9]+)?$/)
```

The `\.[0-9]+` part is not optional, so a number written in the ordinary scientific form, with no
dot, matches neither this branch nor the integer branch, and ends up in the symbol check:

```
(eq foo 1e10):    Wrong symbol format (1e10)
(eq foo 1.5e+10): OK
```

`Float('1e10')` in Ruby is `10000000000.0`, and the same text is a number in JSON too.

This is not #767, which is about the sign after `e` being required when a dot is present. Making
the sign optional would still leave `1e10` refused, because the dot is the part that is missing.

## What should happen

An integer mantissa with an exponent should parse as a float, so that `1e10` means what it means
everywhere else.

morphqdd:
The float pattern required a dot, so `1e10` matched neither the integer branch nor the float one
and was refused as a symbol. An integer mantissa with an exponent parses as a float now.

Closes #1010


yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12314](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

